PR #13, opened 2017-10-11
Expand VersionRegistry with some ProductVersions

tremes:


tremes:
Actually I think this should be expanded more to have an option to get a version of every product.

PR #17, opened 2017-11-03
Snapshot release configuration

dsimansk:
This adds config of snapshot repo and Travis command to release from master.


dsimansk:
@maschmid this should be also ready to merge.

PR #23, opened 2017-11-15
Fix removeRoleFromEntity method

Crumby:
This method seems to get broken with fabric client version update

### DIFF
--- a/src/main/java/cz/xtf/openshift/OpenshiftUtil.java
+++ b/src/main/java/cz/xtf/openshift/OpenshiftUtil.java
@@ -873,11 +873,13 @@ public class OpenshiftUtil implements AutoCloseable {
 
 	private void removeRoleFromEntity(String namespace, String role, String entityKind, String entityName, String userName) {
 		withAdminUser(client -> {
-			if (client.inNamespace(namespace).roleBindings().withName(role).get() != null) {
-				client.inNamespace(namespace).roleBindings().withName(role).edit()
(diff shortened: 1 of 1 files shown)


src/main/java/cz/xtf/openshift/OpenshiftUtil.java
```diff
@@ -873,11 +873,13 @@ public void removeRoleFromServiceAccount(String namespace, String role, String s
 
 	private void removeRoleFromEntity(String namespace, String role, String entityKind, String entityName, String userName) {
 		withAdminUser(client -> {
-			if (client.inNamespace(namespace).roleBindings().withName(role).get() != null) {
-				client.inNamespace(namespace).roleBindings().withName(role).edit()
-								.removeFromSubjects(new ObjectReferenceBuilder().withKind(entityKind).withName(entityName).build())
-								.removeFromUserNames(userName)
-								.done();
+			RoleBinding roleBinding = client.inNamespace(namespace).roleBindings().withName(role).get();
```
dsimansk: Have you considered something like ...ifPresent(rb -> rb.doMoreFunHere()) ? :)

Crumby: .withName(role) return type isn't Optional<RoleBinding> but Resource<RoleBinding, DoneableRolebinding>.

dsimansk: Ehm, sorry my bad :)

PR #32, opened 2018-01-02
OpenShiftUtil reimplemented

Crumby:
This utility re-implements OpenshiftUtil (note the new uppercase S). It's no longer singleton instance driven by properties but rather instantiable class with that is configured once only during instantiation. Thus no more need for OpenShiftContext class and context switching. OpenShiftContext and original Openshift util are now deprecated. Also bunch of code that didn't follow KISS rule has been removed.

### DIFF
--- a/src/main/java/cz/xtf/openshift/OpenShiftUtil.java
+++ b/src/main/java/cz/xtf/openshift/OpenShiftUtil.java
@@ -1,0 +1,869 @@
+package cz.xtf.openshift;
+
+import io.fabric8.kubernetes.api.model.*;
+import io.fabric8.kubernetes.api.model.Service;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.kubernetes.client.dsl.KubernetesListMixedOperation;
+import io.fabric8.kubernetes.client.dsl.LogWatch;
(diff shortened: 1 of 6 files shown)


src/main/java/cz/xtf/openshift/OpenShiftUtil.java
```diff
@@ -0,0 +1,869 @@
...
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+
+@Slf4j
+public class OpenShiftUtil  implements AutoCloseable {
+	private final NamespacedOpenShiftClient client;
+	private final String namespace;
+
+	public OpenShiftUtil(String masterUrl, String namespace, String username, String password) throws MalformedURLException {
```
dsimansk: @Crumby could we add one more constructor to pass OpenShiftConfig object directly. E.g. I'd like to benefit form autoload of `~/.kube/config` or secrets in pod.

Crumby: I have added the possibility, however this config should have set namespace since the utility heavily depends on that.

dsimansk: Sure, works for me :) Thanks ☃️

PR #41, opened 2018-01-18
Waiting with new OpenShiftUtil

Crumby:


src/main/java/cz/xtf/openshift/OpenShiftUtil.java
```diff
@@ -828,4 +773,97 @@ private Path storeLog(String log, Path dirPath, String fileName) throws IOExcept
...
+				boolean noBuild = getBuilds().isEmpty();
+				boolean noRoute = getRoutes().isEmpty();
+				boolean noPod = getPods().isEmpty();
+				boolean noPvc = getPersistentVolumeClaims().isEmpty();
+				boolean noHpa = getHorizontalPodAutoscalers().isEmpty();
+				boolean noCm = getConfigMaps().isEmpty();
+				boolean noUserSecrets = getUserSecrets().isEmpty();
+				boolean noUserSa = getUserServiceAccounts().isEmpty();
+
+				return noDc && noRc && noBc && noIs && noEp && noSvc && noBuild && noRoute && noPod && noPvc && noHpa && noCm && noUserSecrets && noUserSa;
```
dsimansk: Maybe put those booleans into array and evaluate at once? :)

Crumby: Good point, updated.

src/main/java/cz/xtf/wait/Waiter.java
```diff
@@ -0,0 +1,133 @@
...
+	Waiter logPoint(LogPoint logPoint);
+
+	/**
+	 * Waits till condition is met.
+	 *
+	 * @return true if wanted condition was met, false if unwanted state condition was met
+	 * @throws TimeoutException in case that neither wanted nor unwated condition have been met
+	 * @throws InterruptedException by {@link Thread#sleep}
+	 */
+	boolean run() throws TimeoutException, InterruptedException;
```
Crumby: I'm not exactly sure by naming this method. 'wait' would be ideal but that is not an option because of Object.wait(). Another options that came to my mind are: doWait, waitFor, run, execute or start. Ideas?

src/main/java/cz/xtf/openshift/OpenShiftUtil.java
```diff
@@ -828,4 +786,126 @@ private Path storeLog(String log, Path dirPath, String fileName) throws IOExcept
...
+		}
+
+		/**
+		 * Creates a waiter that checks whether all deployment pods are ready.
+		 * Defaults to 2 minutes timeout.
+		 *
+		 * @param dcName name of deploymentConfig
+		 * @return Waiter instance
+		 */
+		public Waiter deploymentReadiness(String dcName) {
```
Crumby: Thinking whether this naming is fluent enough. Typical usage then would be: 

openshift.waiters().deploymentReadiness(args...).run();

Suggestions?

dsimansk: Maybe something more descriptive, e.g.

```
openshift.waiters().isReady().dc(arg).run();
// where isReady() defines condition
// dc() is self-explaining :)
```
or
```
openshift.waiters().isReady().withResource(arg).run();
````
or 
```
openshift.waiters().isReady().resource(arg).run();
```

File: src/main/java/cz/xtf/wait/Waiter.java (hunk 2 above)
Comment (dsimansk) [reply src/main/java/cz/xtf/wait/Waiter.java]: My subjective priority list would be `execute(), doWait(), perform()`.

File: src/main/java/cz/xtf/wait/SupplierWaiter.java [hunk 4]
```diff
@@ -0,0 +1,94 @@
+package cz.xtf.wait;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+public class SupplierWaiter<X> implements Waiter {
```
dsimansk: Imho `WaiterSupplier` sounds a bit better or?

src/main/java/cz/xtf/openshift/OpenShiftUtil.java
```diff
@@ -828,4 +786,126 @@ private Path storeLog(String log, Path dirPath, String fileName) throws IOExcept
...
 	}
+
+	private OpenShiftWaiters waiters = new OpenShiftWaiters();
+
+	// Waiting
+	public OpenShiftWaiters waiters() {
+		return waiters;
+	}
+
+	public class OpenShiftWaiters {
```
dsimansk: Could it be in separate file? This file is long enough by itself :)

src/main/java/cz/xtf/wait/Waiter.java
dsimansk: And maybe I'd like to see `InterruptedException` wrapped with `TimeoutException` for simplifying the declaration. The `catch{}` block was usually shared for both of those, or at least I've done it.:)

Crumby: Wrapping `InterruptedException` seems reasonable.

src/main/java/cz/xtf/wait/SupplierWaiter.java
Crumby: `SupplierWaiter` - `Waiter` that uses `Supplier<X>` to retrieve state of object and those objects are tested for some state by `Function<X, boolean>`
`WaiterSupplier` - class that supplies waiters?

`SupplierWaiter` in my opinion better describes what to expected from class. Like, it's not an `Supplier<>` in java lambda manner.

dsimansk: Ok, I do agree it might be more confusing with other `*Suppliers`. Consider this as resolved then :)

src/main/java/cz/xtf/wait/Waiter.java
Crumby: renamed method to `execute` and swallowed up `InterruptedException`. Note that `assertEventually` is also present if you want to swallow up also TimeoutException.

src/main/java/cz/xtf/openshift/OpenShiftUtil.java
Crumby: Class is now separate.

src/main/java/cz/xtf/openshift/OpenShiftUtil.java
Crumby: Renamed `Waiter` creating methods so they start with either with is, has or does(eg. `doesUrlReturnsOK`). Ability to specify request with like `isReady()` could be good once we have more then few default waiters.

Crumby:
I have hopefully resolved all your notes. If you don't have any other for now I will merge this and start testing it's practical usage with some suite. This could reveal some flaws or needed updates.

I have also created three versions of `openshift.clean()` method. One that only cleans, other that clean and waits and throws `TimeoutException` and one that cleans and asserts cleaned. The last one could be good for testing purposes. Having one that only cleans and asserts by default didn't seem like good idea.

src/main/java/cz/xtf/wait/Waiters.java
```diff
@@ -0,0 +1,67 @@
...
+		try {
+			logPoint.logStart(reason, millis);
+			Thread.sleep(millis);
+			logPoint.logEnd(reason, millis);
+		} catch (InterruptedException e) {
+			log.warn("Interrupted during waiting. Wait reason: {}", reason, e);
+		}
+	}
+
+	public static Waiter doesUrlReturnsOK(String url) {
```
tomason: Grammar rant: doesUrlReturnsOK -> doesUrlReturnOK

src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	/**
+	 * Creates a waiter that checks that exactly n pods is ready in project.
+	 * Defaults to 2 minutes timeout.
+	 *
+	 * @param n number of expected pods to wait upon
+	 * @return Waiter instance
+	 */
+	public Waiter isExactlyNPodsReady(int n) {
```
tomason: Grammar rant: isExactlyNPodsReady -> areExactlyNPodsReady

src/main/java/cz/xtf/wait/Waiters.java
Crumby: Fixed

src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
Crumby: Fixed.

tomason:
Thanks for consideration, grammar looks good to me now 👍

Crumby:
@tomason @dsimansk Thank you for your comments. Pull request is now merged at master.

PR #42, opened 2018-01-23
Delete unused products enum

Crumby:


llowinge:
So should i add the Product enum directly into test-common (your team probably doesn't use it).

Crumby:
Oh, sorry for the trouble, I didn't realize that. Our team no longer uses Product enum anywhere in the suites. AbstractTemplateTest and all the child test where re-factored in October/November. If you want to keep AbstractTemplateTest as is, then for future you'll have to add the enum into the test-common suite

For now, I have reverted this commit and deprecate the enum instead (see https://github.com/xtf-cz/xtf/pull/43).

llowinge:
Ok thank you for info.

PR #46, opened 2018-01-30
Logging and clean up fixes

Crumby:


src/main/java/cz/xtf/openshift/OpenShiftUtil.java
```diff
@@ -523,6 +532,10 @@ public RoleBinding getRoleBinding(String name) {
 		return client.roleBindings().list().getItems();
 	}
 
+	public List<RoleBinding> getUserRoleBindings() {
```
maschmid: needs javadoc to describe what the "user" role bindings are supposed to mean

Crumby: extended with javadoc

PR #53, opened 2018-03-28
[CLOUD-2261] adding roles builder as a way to define role when needed

ochaloup:
While working on fix for https://issues.jboss.org/browse/CLOUD-2261 I would need to have chance to add roles and role binding as the image I use needs to list pods via api call.

Now I would like to add this ability to the xtf for me being able to use it then in xpaas qe test `test-eap/PostgreSQLXARecoveryLoadTest`. This is maintained by @tremes but I do some tweaks in it. I'm aware of the fact the changes goes to the deprecated part of the code (`OpenshiftUtil`) but the whole test is written in way of using it.

If you guide me how to do it in better way I will be more than happy.


dsimansk:
@Crumby fancy doing a review?

PR #55, opened 2018-04-11
Localdev refactoring test configuration

YegorMaksymchuk:
Divide TestConfiguration.Create two separate classes for use XTFConfiguration in radanalytics-utils.


src/main/java/cz/xtf/XTFConfiguration.java
```diff
@@ -0,0 +1,798 @@
...
+	private static final String NFS_SSH_USERNAME = "xtf.config.nfs.ssh_username";
+	private static final String PROXY_HOST = "xtf.config.proxy.host";
+	private static final String PROXY_HOST_USERNAME = "xtf.config.proxy.host.username";
+	private static final String PROXY_HOSTS = "xtf.config.proxy.hosts";
+	private static final String PROXY_SQUID_PORT = "xtf.config.proxy.squid.port";
+	private static final String PROXY_VERTX_PORT = "xtf.config.proxy.vertx.port";
+	private static final String ROUTE_DOMAIN = "xtf.config.route_domain";
+	private static final String PROXY_DOMAIN = "xtf.config.domain.proxy";
+	private static final String SOAK_TEST_ITERATIONS = "xtf.config.soak_test_iterations";
+	private static final String IMAGE_REPO_BASE = "xtf.config.image.repo";
```
zroubalik: this line could be deleted

src/main/java/cz/xtf/XTFConfiguration.java
```diff
@@ -0,0 +1,798 @@
...
+	private static final String TEMPLATE_REPO = "xtf.config.template.repo";
+	private static final String TEMPLATE_BRANCH = "xtf.config.template.branch";
+
+	private static final String OPENSTACK_URL = "xtf.config.openstack.url";
+	private static final String OPENSTACK_TENANT = "xtf.config.openstack.tenant";
+	private static final String OPENSTACK_OPEN_SECURITY_GROUP = "xtf.config.openstack.security.group.open";
+	private static final String OPENSTACK_USERNAME = "xtf.config.openstack.username";
+	private static final String OPENSTACK_PASSWORD = "xtf.config.openstack.password";
+
+	public static final String VERSION_EAP = "xtf.version.eap";
```
zroubalik: VERSION_EAP .... VERSION_MSA should be in the TestConfiguration.java

src/main/java/cz/xtf/TestConfiguration.java
```diff
@@ -479,140 +86,13 @@ public static String getMsaVersion() {
...
-		return readValue(key, null);
-	}
-
-	public String readValue(final String key, final String defaultValue) {
-		return this.properties.getProperty(key, defaultValue);
-	}
-
-	private Properties fromPath(final String path) {
+	@Override
+	protected Properties fromEnvironment() {
```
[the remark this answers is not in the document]
Crumby: Told a little bit differently, there is no TestConfiguration instance that would initialize its fromEnvironment() properties and thus calling something like TestConfiguration.get().readValue(TestConfiguration.IMAGE_EAP) will fail with null pointer exception.

src/main/java/cz/xtf/XTFConfiguration.java
```diff
@@ -0,0 +1,744 @@
...
+	private static final String TEMPLATE_BRANCH = "xtf.config.template.branch";
+
+	private static final String OPENSTACK_URL = "xtf.config.openstack.url";
+	private static final String OPENSTACK_TENANT = "xtf.config.openstack.tenant";
+	private static final String OPENSTACK_OPEN_SECURITY_GROUP = "xtf.config.openstack.security.group.open";
+	private static final String OPENSTACK_USERNAME = "xtf.config.openstack.username";
+	private static final String OPENSTACK_PASSWORD = "xtf.config.openstack.password";
+
+
+	public static final String OC_BINARY_LOCATION = "oc.binary.location";
```
Crumby: Why did this property moved to XTFConfiguration while it's static method stood in TestConfiguration?

YegorMaksymchuk: XTFConfiguration should be like a base configuration class.
So I think it is a common attribute.

src/main/java/cz/xtf/TestConfiguration.java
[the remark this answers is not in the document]
YegorMaksymchuk: Yep. thanks!

src/main/java/cz/xtf/XTFConfiguration.java
Crumby: Shouldn't then also method TestConfiguration.ocBinaryLocation() be rather in XTFConfiguration where it's property lies? Properties and methods should lie within same file.

https://github.com/xtf-cz/xtf/pull/55/files#diff-3cc59395e886593785f8f1ac99a1ba5cR73

src/main/java/cz/xtf/TestConfiguration.java
```diff
@@ -58,6 +58,8 @@
 	public static final String CI_USERNAME = "ci.username";
 	public static final String CI_PASSWORD = "ci.password";
 
+	public static final String OC_BINARY_LOCATION = "oc.binary.location";
```
zroubalik: OC_BINARY_LOCATION with dependent methods should be in XTFConfiguration class, it is general purpose property

YegorMaksymchuk: fixed

src/main/java/cz/xtf/TestConfiguration.java
```diff
@@ -125,494 +44,229 @@
...
-			copyValues(fromPath("global-test.properties"));
-		}
-		else if (Paths.get("../global-test.properties").toAbsolutePath().toFile().exists()) {
-			copyValues(fromPath("../global-test.properties"));
-		}
+	private static final TestConfiguration INSTANCE = new TestConfiguration();
 
-		// then defaults
-		copyValues(defaultValues());
+	protected TestConfiguration() {
```
Crumby: This class shouldn't be netiher instantiatable  at all nor it shouldn't hold any properties. Those are actually copied in `XTFConfiguration` class and there is no need to hold pointer also in this class. Constructor should be marked as private and this class properties instantiation should be moved to static block:

`static
{
get().copyValues(fromEnvironment());
get().copyValues(defaultValues());
}`

after that something like this should work without problem:
`get().readValue(TestConfiguration.IMAGE_MSA)`

src/test/java/cz/xtf/configuration/TestConfigurationTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class TestConfigurationTest {
+
+	@Test
+	public void canReadDefaultValueOfPropertyFromXTFConfiguration() {
+		Assert.assertEquals("/usr/bin/oc", TestConfiguration.ocBinaryLocation());
+	}
+
+	@Test
+	public void canReadDefaultValueByTestConfigurationFromXTFConfiguration() {
```
Crumby: I don't see difference between the tests.

src/main/java/cz/xtf/XTFConfiguration.java
```diff
@@ -0,0 +1,751 @@
...
+	private static final String OPENSTACK_USERNAME = "xtf.config.openstack.username";
+	private static final String OPENSTACK_PASSWORD = "xtf.config.openstack.password";
+
+	public static final String OC_BINARY_LOCATION = "oc.binary.location";
+
+
+	public static final String CDK_DOMAIN = "cd.xtf.cz";
+	public static final String CDK_IP = "10.1.2.2";
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(XTFConfiguration.class);
```
Crumby: Could you please move to using `@Slf4j` annotation while at refactoring?

src/main/java/cz/xtf/TestConfiguration.java
```diff
@@ -109,510 +35,231 @@
...
 	public static final String CDK_INTERNAL_HOSTNAME = "localhost.localdomain";
-
-	public static final String OC_BINARY_LOCATION = "oc.binary.location";
-
 	public static final String CI_USERNAME = "ci.username";
 	public static final String CI_PASSWORD = "ci.password";
 
-	private static final Logger LOGGER = LoggerFactory.getLogger(TestConfiguration.class);
+	private Properties properties;
 	private static final TestConfiguration INSTANCE = new TestConfiguration();
```
Crumby: Same as above. No need to hold INSTANCE. All the properties are held in XTFConfiguration.

src/main/java/cz/xtf/TestConfiguration.java
```diff
@@ -109,510 +35,231 @@
...
-
-	public static String openStackPassword() {
-		return get().readValue(OPENSTACK_PASSWORD);
+	private static String getProperty(String property) {
+		return getTC().properties.getProperty(property);
 	}
 
 	public static String ciUsername() {
-		return get().readValue(CI_USERNAME);
+		return getProperty(CI_USERNAME);
```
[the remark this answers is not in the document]
YegorMaksymchuk: I wrapped get().readValue() in method getProperty();

src/main/java/cz/xtf/TestConfiguration.java
```diff
@@ -109,510 +35,231 @@
...
-
-		return props;
+	public static String versionMsa() {
+		return getProperty(VERSION_MSA);
 	}
 
-	private Properties fromEnvironment() {
-		final Properties props = new Properties();
+	protected static Properties fromEnvironment() {
+		final Properties props = get().fromEnvironment();
```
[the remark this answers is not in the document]
YegorMaksymchuk: fixed

PR #57, opened 2018-04-12
CLOUD-2455 _TRACKING RA attribute

maschmid:


src/main/java/cz/xtf/openshift/messaging/JBossAMQ.java
```diff
@@ -100,6 +101,11 @@ public void configureApplicationDeployment(final DeploymentConfigBuilder dcBuild
 				.envVar(getEnvVarName("TOPICS"), getTopicList())
 				.envVar(getEnvVarName("PROTOCOL"), "tcp");
 
+		if (tracking != null) {
+			dcBuilder.podTemplate().container()
+					.envVar(getEnvVarName("TRACKING"), tracking.booleanValue() ? "true" : "false");
```
Crumby: if(true) true else false? tracking.booleanValue() should be enough. Or why don't we initialize tracking with false by default?

maschmid: it shouldn't be false by default. it should be the default by default (ie, empty)

maschmid: the fact that Java boolean toString is 'true'/'false' is just a coincidence that it has the same format that the env file accepts, so I'd like to have the mapping between the boolean and the string explicit.

PR #64, opened 2018-04-25
adding areNoPodsPresent and isLatestBuildPresent Waiters

zroubalik:


src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
```diff
@@ -28,12 +28,44 @@
...
+
+	/**
+	 * Creates waiter for latest build presence with default timeout
+	 * 5 seconds interval check and both logging points.
+	 *
+	 * @param buildConfigName name of buildConfig for which build to be waited upon
+	 * @param buildNamespace namespace containing the buildConfig
+	 * @return Waiter instance
+	 */
+	public static Waiter isLatestBuildPresent(String buildConfigName, String buildNamespace) {
```
Crumby: OpenShiftWaiters shouldn't parameterize buildNamespace... it's namespace is parameterized by OpenShiftUtil. So if you want Waiter with different namespace you should do something like:

`OpenShiftUtil.master(buildNamespace).waiters().isLatestBuildPresent(buildConfigName);`

src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
```diff
@@ -28,12 +28,44 @@
...
+		Supplier<Build> supplier;
+
+		if (buildNamespace == null || buildNamespace.isEmpty()) {
+			supplier = () -> OpenShiftUtils.master().getLatestBuild(buildConfigName);
```
Crumby: Use openShiftUtil field.

src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
```diff
@@ -190,4 +222,23 @@ public Waiter areExactlyNPodsRunning(int n, String key, String value) {
...
+
+	/**
+	 * Creates a waiter that waits until there aren't any pods in project.
+	 * Defaults to 3 minutes timeout.
+	 *
+	 * @param key label key for pod filtering
+	 * @param value label value for pod filtering
+	 * @return Waiter instance
+	 */
+	public static Waiter areNoPodsPresent(String key, String value) {
```
Crumby: Isn't following use case sufficient for this purpose?

`areExactlyNPodsRunning(0, key, value)`

src/main/java/cz/xtf/openshift/ResourceFunctions.java
```diff
@@ -23,6 +23,10 @@
 		return pods -> pods.stream().anyMatch(pod -> ResourceParsers.hasPodRestartedAtLeastNTimes(pod, n));
 	}
 
+	public static Function<List<Pod>, Boolean> areNoPodsPresent() {
```
Crumby: This is way to trivial to be here... Also you can achieve the same result by using either `areExactlyNPodsReady(0)` or `areExactlyNPodsRunning(0)`.

src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
zroubalik: Unfortunately not, need to be sure there aren't any pods. Your proposed solution, doesn't cover other Pod's state (Pending, Failed,...)

src/main/java/cz/xtf/openshift/ResourceFunctions.java
zroubalik: Same as above

src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
zroubalik: Good catch, you are right. Was using these waiter methods in another helper class and I copypasted them here without thinking.

src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
Crumby: Rethinking that and you are correct. It would actually work if you look at the implementation, but looking at the implementation, it's not exactly correct. Also, yep, from method name you could expect different behavior.

src/main/java/cz/xtf/openshift/ResourceFunctions.java
Crumby: Correct. Still this method shouldn't be here since it's way to trivial and only says whether list size is zero. It doesn't even work with resource itself (Could be be `List<X>` instead `List<Pod>`). Also what about `pods.isEmpty()`?

src/main/java/cz/xtf/openshift/OpenShiftWaiters.java
```diff
@@ -190,4 +222,23 @@ public Waiter areExactlyNPodsRunning(int n, String key, String value) {
...
+		Supplier<List<Pod>> ps = () -> OpenShiftUtils.master().getLabeledPods(key, value);
+		String reason = "Waiting for no present pods with label " + key + "=" + value + ".";
+
+		return areNoPodsPresent(ps).reason(reason);
+	}
+
+	private static Waiter areNoPodsPresent(Supplier<List<Pod>> podSupplier) {
+		return new SupplierWaiter<>(podSupplier, ResourceFunctions.areNoPodsPresent(), TimeUnit.MINUTES, 3);
```
Crumby: `return new SupplierWaiter<>(podSupplier, pods -> pods.isEmpty(), TimeUnit.MINUTES, 3);`

Crumby: Or even better:
`return new SupplierWaiter<>(podSupplier, List::isEmpty, TimeUnit.MINUTES, 3);`

PR #69, opened 2018-05-02
reloaded hasBuildCompleted methode with timeout parameter

olegvn88:


Crumby:
This breaks intended way of using Waiters... The main reason is not to have a bunch of overriding methods with bunch of parameters. So if you don't like default wait timeout just call eg.: `hasBuildCompleted(buildName).timeout(TimeUnit timeUnit, long timeout).execute();`

olegvn88:
Thanks, I wasn't aware of that possibility

PR #75, opened 2018-05-16
OpenShiftUtil - delete also roles within clean

tremes:


Crumby:
Please update OpenShiftWaiters.isProjectClean() method so it also considers roles.

PR #82, opened 2018-06-03
configurable ssh channel connection timeout

maschmid:


maschmid:
scratch this, the default 20 seconds should be enough for everybody.

PR #84, opened 2018-06-08
[fix] Allow concurrent invocation

llowinge:


Crumby:
@llowinge Trying to figure out what is happaning here... So we delete project, client will return 'null' and we try to create new one getting error that project already exist... Are we sure that project was recreated in case `recreateIfExist=true` and project won't get deleted few seconds after?

llowinge:
I was trying to fix scenario, where more instances of OpenshiftUtils can call this method in same time and race condition can happen in the way that first one will create the new namespace and the second one in same time will also think that it doesn't exist so it attempts to create the new namespace also and it fails - but it fails only in case, that the namespace already doesn't exists from beginning - so i think i can overcome it when i create the desired namespace (i use the same in all cases) manually, then all of them will skip the creation from the beginning...so for me i found workaround now and i don't need this PR anymore.

dsimansk:
@llowinge fancy adding an exists check that @Crumby requested? All you'll need to rebase eventually since master was modularized meanwhile.

llowinge:
@Crumby  Do you mean adding additional check into if statement - "&& getProject(name) != null" ?

Crumby:
@llowinge exactly.

llowinge:
@Crumby  Fixed

PR #85, opened 2018-06-11
Extended clean() to delete statefulsets

penehyba:


src/main/java/cz/xtf/openshift/OpenShiftUtil.java
```diff
@@ -883,6 +883,7 @@ public void cleanAndAssert() {
 	public void clean() {
 		// keep the order for deletion to prevent K8s creating resources again
 		client.extensions().deployments().delete();
+		client.apps().statefulSets().delete();
```
Crumby: You should also extend `OpenShiftWaiters::isProjectClean` so it checks stateful sets to be deleted

penehyba: Done!

PR #95, opened 2018-06-18
JBossAMQ - add option to pass the service name

tremes:


maschmid:
we should probably add withServiceName then in addition to (or instead of? ) ctr.

tremes:
Could be...this option came to my mind as well. So probably instead of constructor.

PR #100, opened 2018-06-20
[http-client] HttpClient module

Crumby:


dsimansk:
I wonder if we shouldn't also upgrade dependency versions while iterating over util classes and moving them to new home. We're on `4.3.6`, but the latest is `4.5.5`. We'll hit some API compatibility issue, no doubts about that :)

PR #101, opened 2018-06-22
JBossAMQ - add option to use AMQ7 image

tremes:


tremes:
It would be better to introduce new `JBossAMQ7` class.

PR #103, opened 2018-06-26
[core] add OpenShift::portForward method

dsimansk:
Simple `PodPortForward` impl without `oc` binary.

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -4,6 +4,7 @@ import cz.xtf.core.waiting.SimpleWaiter;
 import cz.xtf.core.waiting.Waiter;
 import io.fabric8.kubernetes.api.model.*;
 import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.kubernetes.client.LocalPortForward;
 import io.fabric8.kubernetes.client.dsl.LogWatch;
 import io.fabric8.openshift.api.model.*;
 import io.fabric8.openshift.client.*;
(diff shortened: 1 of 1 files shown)


core/src/main/java/cz/xtf/core/openshift/PodPortForward.java
```diff
@@ -0,0 +1,55 @@
...
+	private String podName;
+
+	private LocalPortForward connection;
+
+	public PodPortForward(OpenShift openShift, Pod pod) {
+		this.openShift = openShift;
+		this.podName = pod.getMetadata().getName();
+	}
+
+	public PodPortForward(OpenShift openShift, String dcName) {
```
Crumby: Could we switch constructors order? I consider it more readable if called constructor (method) is bellow instead above.

core/src/main/java/cz/xtf/core/openshift/PodPortForward.java
```diff
@@ -0,0 +1,55 @@
+package cz.xtf.core.openshift;
+
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.client.LocalPortForward;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class PodPortForward implements AutoCloseable {
+
```
Crumby: Could we delete this empty line? Just to keep it consistent with other classes.

Crumby:
What is an actual advantage of this class, in the end usage meaning. Autoclosability and accessibility with just dcName?

core/src/main/java/cz/xtf/core/openshift/PodPortForward.java
dsimansk: Sure, we probably need a codestyle check then :)

dsimansk:
@Crumby the motivation is to keep AutoCloseable that we used to be provided via `PodService`. There's no additional value actually. Maybe, we could extend it a bit more to reconnect on failure? Would that make more sense?

core/src/main/java/cz/xtf/core/openshift/PodPortForward.java
Crumby: Created an issue https://github.com/xtf-cz/xtf/issues/104 for that.

PR #106, opened 2018-06-27
master url is more specific setting than DOMAIN

maschmid:


maschmid:
closing in favor of #109

Issue #104, opened 2018-06-27
Consider checkstyle implementation


Crumby:
We should take a look at checkstyle possibilities to keep all classes consistent.

dsimansk:
I've added a checkstyle via #108 with a couple of rules we used to have on previous utils iterations. However, we could even consider more standard rulesets, like Google's or Sun's.

https://github.com/checkstyle/checkstyle/tree/master/src/main/resources
http://checkstyle.sourceforge.net/google_style.html
http://checkstyle.sourceforge.net/sun_style.html

mnovak1:
@Crumby what about to use maven formatter plugin [1] which automatically format the code during build?

[1] https://code.revelc.net/formatter-maven-plugin/

mnovak1:
Automatic formatting is done in: https://github.com/xtf-cz/xtf/pull/394

PR #108, opened 2018-06-28
[parent] add checkstyle

dsimansk:
Related to #104


checkstyle.xml
```diff
@@ -0,0 +1,34 @@
...
+		<!-- Force tab indentation -->
+		<module name="RegexpSinglelineJava">
+			<property name="format" value="^\t* "/>
+			<property name="message" value="Indent must use tab characters"/>
+			<property name="ignoreComments" value="true"/>
+		</module>
+
+		<module name="OverloadMethodsDeclarationOrder"/>
+
+		<!--<module name="AvoidStarImport"/>-->
```
Crumby: I would remove this line to keep it as clear as we can.

dsimansk: Yep, I wanted to discuss the rule actually :) I'd prefer no star imports personally.

dsimansk:
> Does checkstyle executes automatically or do we need to add it as callback to project settings?

It'll now. I forgot to add `execution` part.

checkstyle.xml
Crumby: Fine by me... Even might be a little bit annoying with Idea that automatically uses star imports. But that's configurable I believe.

PR #117, opened 2018-07-04
[core] a pod can be ready only if it has at least one container

maschmid:
a possible explanation for the "container not found" race


maschmid:
I have verified that it is indeed the case that 

openShift.waiters().isDcReady(appName).waitForOrAssertFail("Failed while waiting for dc to get ready");

sometimes immediately returns with a pod that doesn't have containers yet.

PR #121, opened 2018-07-11
added method createForPod with host argument

olegvn88:
Added createForPod method with host argument, because of on some Openshift environments pod’s spec doesn’t contain hostname of the node

### DIFF
--- a/utilities/src/main/java/cz/xtf/docker/DockerContainer.java
+++ b/utilities/src/main/java/cz/xtf/docker/DockerContainer.java
@@ -1,6 +1,13 @@
 package cz.xtf.docker;
 
-import static cz.xtf.openshift.OpenShiftUtils.admin;
+import cz.xtf.TestConfiguration;
+import cz.xtf.openshift.OpenshiftUtil;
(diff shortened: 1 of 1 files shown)


utilities/src/main/java/cz/xtf/docker/DockerContainer.java
```diff
@@ -63,9 +61,13 @@ public static DockerContainer createForPod(Pod pod) {
...
 	}
 
+
 	public static DockerContainer createForPod(Pod pod, String containerLabel) {
 		String host = pod.getSpec().getNodeName();
+		return createForPod(pod, containerLabel, host);
+	}
 
+	public static DockerContainer createForPod(Pod pod, String containerLabel, String host) {
 		try {
```
maschmid: this part (the whole try-catch block with guessing the hostname from nodeName should be moved to the createForPod(Pod pod, String containerLabel)  

if explicit host is provided, it should just be used as a host, with no additional hacks.

olegvn88: Done.

PR #135, opened 2018-08-08
[fix] Use higher timeout for amq builds

llowinge:


dsimansk:
@llowinge if it's causing an issue. Let's make it configurable, e.g `withTimeout()` method and property to the class. 5 min was be pretty sufficient for AMQ over the years :), even if it's pulled to node for the first time.

llowinge:
@dsimansk Is it OK now ?

PR #136, opened 2018-08-08
[utilities] Use openssl to retreive ocp certs

Crumby:


utilities/src/main/java/cz/xtf/keystore/ProcessKeystoreGenerator.java
```diff
@@ -38,11 +34,14 @@
...
-			if(!TestConfiguration.openshiftOnline()) {
-				String routerCrt = OpenShiftNode.master().executeCommand("sudo cat /etc/origin/master/ca.crt");
-				FileUtils.writeStringToFile(caDir.resolve("openshift-router.pem").toFile(), routerCrt);
-				processCall(caDir, "keytool", "-import", "-noprompt", "-keystore", "truststore", "-file", "openshift-router.pem", "-alias", "openshift-router", "-storepass", "password");
-			}
+			processCall(caDir, "/bin/sh", "-c", "echo \"Q\" | openssl s_client -connect " + TestConfiguration.masterUrl().replaceFirst("https://", "") + " -showcerts 2>/dev/null > serversOpenSslResponse");
+			processCall(caDir, "/bin/sh", "-c", "csplit -f serverCert -s serversOpenSslResponse '/^-----BEGIN CERTIFICATE-----$/' '{*}'");
+
+			processCall(caDir, "openssl", "x509", "-in", "serverCert01", "-out", "cert01.pem");
+			processCall(caDir, "openssl", "x509", "-in", "serverCert02", "-out", "cert02.pem");
```
maschmid: wouldn't it be better to not assume there are always just two certs in the chain?

PR #142, opened 2018-08-13
[core] BuildManager implementation

Crumby:


core/src/main/java/cz/xtf/core/bm/BuildManagers.java
```diff
@@ -0,0 +1,42 @@
...
+import io.fabric8.kubernetes.api.model.ResourceQuotaBuilder;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class BuildManagers {
+	private static BuildManager bm;
+
+	public static BuildManager get() {
+		if (bm == null) {
```
maschmid: we should make it thread-safe  (using double-checked locking)

PR #146, opened 2018-08-22
[utilities] fix getDeploymentConfigEnvVars NPE for not set values of env variable

jakubschwan:
getDeploymentConfigEnvVars(String name) in OpenShiftUtil throws NPE when is collect env. variable without set value. Not set env variable is represented as null.

### DIFF
--- a/utilities/src/main/java/cz/xtf/openshift/OpenShiftUtil.java
+++ b/utilities/src/main/java/cz/xtf/openshift/OpenShiftUtil.java
@@ -433,7 +433,7 @@ public class OpenShiftUtil implements AutoCloseable {
 	 * @return Map of environment variables
 	 */
 	public Map<String, String> getDeploymentConfigEnvVars(String name) {
-		return getDeploymentConfig(name).getSpec().getTemplate().getSpec().getContainers().get(0).getEnv().stream().collect(Collectors.toMap(EnvVar::getName, EnvVar::getValue));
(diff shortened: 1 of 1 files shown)


jakubschwan:
From stack trace NPE is thrown by Collector.toMap, because getValue returns null.

java.lang.NullPointerException
	at java.util.HashMap.merge(HashMap.java:1225)
	at java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1320)
	at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1380)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at cz.xtf.openshift.OpenShiftUtil.getDeploymentConfigEnvVars(OpenShiftUtil.java:436)

maschmid:
You should change it to not use Collectors.toMap then, as it apparently does not allow null values.

Crumby:
Out of date. Utilities are no longer part of the project.

PR #153, opened 2018-09-04
[utilities] secrets creation for registry.redhat.io

maschmid:
Automatically create the pull secrets for an external openshift auth registry (e.g. registry.redhat.io) in the test and build-manager projects.


maschmid:
@dsimansk  @tremes  FYI

PR #173, opened 2018-09-25
[junit5] OpenShift CleanUp extensions

Crumby:


dsimansk:
@Crumby one idea. We might benefit from a parameter such as `cleanAsAdmin`. But we can leave it as an extension for future. Otherwise seems all right to me.

PR #174, opened 2018-10-03
[utilities] Added oracle extension image

penehyba:
Image contains oracle stuff in /extensions/. This enables template with external DB to include external driver thru EXTENSIONS_IMAGE.


utilities/src/main/java/cz/xtf/openshift/imagestream/ImageRegistry.java
```diff
@@ -191,6 +191,10 @@ public String tcpProxy() {
 		return normalize(TestConfiguration.get().readValue(TestConfiguration.IMAGE_TCP_PROXY));
 	}
 
+	public String oracleDriver() {
+		return this.normalize(TestConfiguration.get().readValue("xtf.oracle-driver"));
```
Crumby: This breaks consistency in approach we are handling the images atm... Please add corresponding values also to `TestConfiguration` class.

PR #197, opened 2018-10-23
[utilities] AmqStandaloneBuilder * withMeshConf()

penehyba:


utilities/src/main/java/cz/xtf/openshift/messaging/AmqStandaloneBuilder.java
```diff
@@ -55,6 +58,9 @@ public AmqStandaloneBuilder(String appName, String podName) {
 		if (StringUtils.isBlank(podName)) {
 			this.podName = appName + "-pod";
 		}
+		this.meshDiscoveryType = "kube";
+		this.meshServiceName = "amq-mesh";
+		this.meshServiceNamespace = OpenshiftUtil.getInstance().getContext().getNamespace();
```
dsimansk: Maybe namespace value from config class would be safer here.

PR #200, opened 2018-10-25
Fixed listZipFilesInDir method to match only .zip files (matched also…

StanleyKaleta:
… .zipnew files before)

### DIFF
--- a/test-helpers/src/main/java/cz/xtf/testhelpers/image/ImageContent.java
+++ b/test-helpers/src/main/java/cz/xtf/testhelpers/image/ImageContent.java
@@ -107,7 +107,7 @@ public class ImageContent {
 	}
 
 	public List<String> listZipFilesInDir(String path) {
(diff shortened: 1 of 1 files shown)


test-helpers/src/main/java/cz/xtf/testhelpers/image/ImageContent.java
```diff
@@ -107,7 +107,8 @@ public PodShell shell() {
 	}
 
 	public List<String> listZipFilesInDir(String path) {
-		return shell.executeWithBash("ls -R1 ~ " + path + " | grep '\\.zip'").getOutputAsList();
+		return shell.executeWithBash("ls -R1 ~ " + path + " | grep '\\.zip'").getOutputAsList()
```
Crumby: `...grep '\\.zip$'...` or `...grep '\\.zip\\$'...` should do the job... we are already filtering results by grep no need to do it again in java. However I'm no sure witch version is correct and whether '$' needs to be escaped.

PR #203, opened 2018-10-26
Fixed listZipFilesInDir method: \$ not working

StanleyKaleta:


dsimansk:
@Crumby fyi merging this one. Seems like no-escape was the right solution.

PR #213, opened 2018-11-29
[core] add method to retrieve authToken

dsimansk:


core/src/main/java/cz/xtf/core/http/Https.java
```diff
@@ -111,16 +116,45 @@ public static String httpsGetContent(URL url) {
...
 	private static HttpsURLConnection getHttpsConnection(URL url) {
+		return getHttpsConnection(url, null, null);
+	}
+
+	private static HttpsURLConnection getHttpsConnection(URL url, String username, String password) {
 		try {
 			SSLContext sslContext = SSLContext.getInstance("TLS");
 			sslContext.init(null, new TrustManager[]{new TrustAllManager()}, new SecureRandom());
-
+
```
Crumby: Empty line with tabs...

dsimansk:
> Why isn't class providing methods with basic auth also for getCode and getContent?
Why getHeaders is only for https protocol?

Well, I guess I'm too lazy to impl all the methods to have uniform API, shame on me. I'm fine either way adding the rest of methods or exposing `getHttpsConnection` and its http counterpart and move the code directly to `getToken` method. The second option is probably even better as I'd not expect headers retrieving method to be super popular these days.

dsimansk:
@Crumby I went for the second option. It look cleaner and as I stated above, headers retrieval aren't widely used anywhere else to be needed in `Https` class. Thanks for suggestion.

PR #228, opened 2019-01-10
[core] ability to override strategy in BinaryBuild

maschmid:
allow override of strategy  (to allow Docker strategy instead of Source)

### DIFF
--- a/core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
+++ b/core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
@@ -50,14 +50,16 @@ import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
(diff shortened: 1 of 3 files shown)


core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -189,11 +197,14 @@ private BuildConfig createBcDefinition() {
 		}
 
 		ObjectMeta metadata = new ObjectMetaBuilder().withName(id).withLabels(Collections.singletonMap(CONTENT_HASH_LABEL_KEY, getContentHash())).build();
-		BuildConfigSpec spec = new BuildConfigSpecBuilder()
+		BuildConfigSpecBuilder bcBuilder = new BuildConfigSpecBuilder();
+		bcBuilder
```
dsimansk: Formatting seems to be a bit misaligned.

PR #248, opened 2019-02-05
[fix] Create secret to custom image stream namespace

llowinge:


utilities/src/main/java/cz/xtf/junit/XTFTestSuite.java
```diff
@@ -217,6 +217,9 @@ private static void createImageStreamsProject() {
 				log.info("action=create-project status=START namespace={}", IS_NAMESPACE);
 				openshift.createProjectRequest(IS_NAMESPACE);
 				log.info("action=create-project status=FINISH namespace={}", IS_NAMESPACE);
+				log.info("action=create-or-reg-secret status=START namespace={}", IS_NAMESPACE);
```
maschmid: "create-oreg-secret" would probably be a better title  "oreg" is the name of the openshift installer property, which presumably means "openshift registry"

llowinge: Fixed. Thanks

PR #250, opened 2019-03-12
[core] return String and fix corrupted STDOUT

Crumby:
* `execute()` now returns String with standard output.
* Fixes `Corrupted STDOUT by directly writing to native stream in forked JVM`

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
@@ -1,10 +1,17 @@
 package cz.xtf.core.openshift;
 
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.ArrayUtils;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
```diff
@@ -31,27 +34,32 @@ public void startBuild(String buildConfig, String sourcePath) {
...
-		pb.redirectError(ProcessBuilder.Redirect.INHERIT);
-
-		int result = -1;
+		pb.redirectOutput(ProcessBuilder.Redirect.PIPE);
+		pb.redirectErrorStream(true);
 
 		try {
-			result = pb.start().waitFor();
+			Process p = pb.start();
+			int result = p.waitFor();
```
maschmid: I suspect this can deadlock, if the process is writing more than the size of the buffer, as you are not consuming its stdout.

maschmid: you should probably read the InputStream in a separate thread instead after start.

PR #255, opened 2019-03-29
[builder] add option to specify client in ApplicationBuilder

dsimansk:
This PR adds a convenient method to pass client ref directly via `buildApplication()`.

```java
new OpenShiftApplication(applicationBuilder, openshift).deploy();
// is the same as
applicationBuilder.buildApplication(openshift).deploy();
```


maschmid:
Personally I think that OpenShiftApplication(ApplicationBuilder appBuilder)  ctor (the one that implicitly calls OpenShifts.master() should not even exist... (perhaps we should at least deprecate it)

PR #257, opened 2019-04-01
[all] upgrade openshift-client to v4.2.2

dsimansk:
Another take on client upgrade.

I wonder if we should implement some tests based on [mock servers](https://github.com/fabric8io/kubernetes-client/tree/master/openshift-server-mock/src/main/java/io/fabric8/openshift/client/server/mock) to be on a safe side. E.g. I haven't spotted any `Endpoints` tests even though the API was broken with double-plural issue in `v4.1.3`.

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -13,8 +13,6 @@ import io.fabric8.kubernetes.api.model.HorizontalPodAutoscaler;
 import io.fabric8.kubernetes.api.model.KubernetesList;
 import io.fabric8.kubernetes.api.model.Node;
 import io.fabric8.kubernetes.api.model.ObjectMetaBuilder;
-import io.fabric8.kubernetes.api.model.ObjectReference;
-import io.fabric8.kubernetes.api.model.ObjectReferenceBuilder;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.ReplicationController;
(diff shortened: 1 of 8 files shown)


Crumby:
@dsimansk I'm not sure if mock server could discover this issue but adding tests can't  cause any harm, can it? Did you try to run some tests against OpenShift with the update?

dsimansk:
@Crumby we actually I'm trying to solve one listing issue right now. Pls do not merge yet. I'll let you know the result.

Crumby:
@dsimansk Is this still WIP?

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -644,96 +643,81 @@ public RoleBinding getRoleBinding(String name) {
...
 		RoleBinding roleBinding = getRoleBinding(name);
 
 		if (roleBinding == null) {
+			// e.g `view` is ClusterRole in K8s RBAC
+			// if the roles isn't present fallback to ClusterRole
+			String kind = rbac().roles().withName(name).get() == null ? "ClusterRole" : "Role";
 			roleBinding = new RoleBindingBuilder()
 					.withNewMetadata().withName(name).endMetadata()
-					.withNewRoleRef().withName(name).endRoleRef()
+					.withNewRoleRef().withKind(kind).withName(name).endRoleRef()
```
dsimansk: @Crumby imo this's ready. However, I'm not particularly sure about this "opinionated" part of code. I seemed to work in my tests, but...

maschmid:
I think we should do a release before merging this one.

pom.xml
```diff
@@ -44,7 +44,7 @@
 
 		<!-- Depedency version properties -->
 		<version.httpclient>4.5.5</version.httpclient>
-		<version.openshift-client>4.1.1</version.openshift-client>
+		<version.openshift-client>4.2.0</version.openshift-client>
```
simkam: 4.2.2 is already out, would be nice to upgrade directly to that version

dsimansk: Sure, that shouldn't be a problem.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
maschmid: maybe make it explicit (have  addClusterRoleToServiceAccount  )

dsimansk: I do agree that'd make the methods cleaner. However, if we split the API to `addRole...` and `addClusterRole...`, it'll eventually break all the usage where `view` is added to SA.

dsimansk:
@Crumby @maschmid 
After further discussion with Marek. I done the following changes.
- Added `addRole*` methods that accept new parameter `String roleKind`
- Changed `getOrCreateRoleBinding` to substitute roleKind for `{admin, edit, view}` that are considered [user-facing roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles)
- Deprecated method `addRoleToGroup` to redirect users to new method with roleKind param
- Didn't add handling of `ClusterRoles` those can be still accessed via Fabric8's API

PR #261, opened 2019-04-10
[fix] Don't cache the hash in memory

llowinge:


Crumby:
@llowinge Can you add a description and purpose of this PR? This will cause re-tarring and recomputing hash every time 'BuildMananger.deploy(ManagedBuild mb)' is called. Need for recomputing hash suggest changes in something that should be final once created. That beats the original reason of having pre-defined shared builds that are not changing once created and deployed.

llowinge:
@Crumby The scenario is, deploy app from BinarySourceBuild and then change locally the source and re-deploy again. The hash is still the same (which then leads to no rebuild of the app through BuildManager as it looks like there are no changes in source code), but it should change, right ?

maschmid:
Do you use BuildManager at all in that scenario, or just BinarySourceBuild directly?

llowinge:
Yes i use it like this
First deploy
```
BuildManager bm = BuildManagers.get();
mb = new BinarySourceBuild(stiImage, project, new HashMap<>(), APP_NAME);
final ManagedBuildReference mbr = bm.deploy(mb);
```

Then change the source and deploy again
`bm.deploy(mb);`

Crumby:
@llowinge `BuildManager`/`BinaryBuild` wasn't designed to maintain dynamically changing builds. And it definitely isn't it's main purpose. What could be done however is to add publid explicit `recomputeHash()` method to invalidate or you can extend `binaryBuild()` and override `needsUpdate()`/`getContentHash()`.

@maschmid Thoughts?

maschmid:
We could introduce a private boolean isCached  to BinaryBuild, and expose it via adding a new class MutableBinarySourceBuild (which would set isCached to false)

llowinge:
Hi @Crumby, yes, i can try. Thx

PR #266, opened 2019-05-02
[fix] Keep oc config together with binary

llowinge:


llowinge:
@Crumby Done.

Crumby:
@llowinge Looks fine. Even though I would rather name the methods with directory instead folder to keep that consistent with Java but that is just a nitpick. I need to resolve failing travis before merging.

llowinge:
@Crumby Thanks for feedback

PR #267, opened 2019-05-03
[feature] Create recreate project method with parametrize timeout

llowinge:


llowinge:
@Crumby Created in https://github.com/xtf-cz/xtf/pull/268

PR #269, opened 2019-05-03
Update lombok version

Crumby:
Updating Lombok version should fix failing travis job with Java 11.


jbliznak:
Just curious, is there anything preventing the update to the latest 1.18.x version directly? From changelog it looks we should be safe to do it https://github.com/rzwitserloot/lombok/blob/master/doc/changelog.markdown

Crumby:
@jbliznak I don't think that there is reason why 1.18.x wouldn't work. I just wanted to do quickest and safest update without diving in possible errors.

PR #270, opened 2019-05-03
[feature] Add MutableBinarySourceBuild

llowinge:


core/src/main/java/cz/xtf/core/bm/MutableBinarySourceBuild.java
```diff
@@ -0,0 +1,15 @@
+package cz.xtf.core.bm;
+
+import java.nio.file.Path;
+import java.util.Map;
+
```
maschmid: this really needs a javadoc explaining its purpose)

llowinge:
@maschmid Done.

PR #271, opened 2019-05-06
[core] Add methods for checking the http response

llowinge:


core/src/main/java/cz/xtf/core/http/Https.java
```diff
@@ -46,6 +49,31 @@ public static Waiter doesUrlReturnCode(String url, int expectedCode, int failCod
 		return new SupplierWaiter<>(getCode, x -> x == expectedCode, x -> x == failCode);
 	}
 
+	public static Waiter doesUrlResponseEqual(String url, String... strings) {
```
Crumby: I'm bit confused by this. In what case can one response be equal to more then one string? Same applies for `StartWith`

llowinge: Yes, you were right. It was copy/paste error.

PR #272, opened 2019-05-06
[feature] Add waiter for response containing string

llowinge:


maschmid:
it should rather do 1 request for N strings (currently it would do N requests for N strings per each wait interval)

llowinge:
@maschmid Fixed

PR #273, opened 2019-05-07
Add way to use token instead of username/password for admin.

mchoma:
There is bug in openshift-client which does not allow currently to
connect with username to OCP 4.x https://github.com/fabric8io/kubernetes-client/issues/1505

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
@@ -42,7 +42,11 @@ public class OpenShifts {
 	}
 
 	public static OpenShift admin(String namespace) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
```diff
@@ -42,7 +42,11 @@ public static OpenShift admin() {
 	}
 
 	public static OpenShift admin(String namespace) {
-		return OpenShift.get(OpenShiftConfig.url(), namespace, OpenShiftConfig.adminUsername(), OpenShiftConfig.adminPassword());
+		if(OpenShiftConfig.token() == null) {
```
Crumby: `adminToken()`?

mchoma: yes

PR #275, opened 2019-05-09
Properties config feature

spriadka:
Added option to specify the `test.properties` and `global-test.properties` as a system property of `config.xtf.test.properties.path` and `config.xtf.global.test.properties.path`

### DIFF
--- a/core/src/main/java/cz/xtf/core/config/XTFConfig.java
+++ b/core/src/main/java/cz/xtf/core/config/XTFConfig.java
@@ -29,11 +29,13 @@ public final class XTFConfig {
 
 	private static final Properties properties = new Properties();
 
+	private static final String TEST_PROPERTIES_PATH = "xtf.test_properties.path";
+	private static final String GLOBAL_TEST_PROPERTIES_PATH = "xtf.global_test_properties.path";
+
 	// Pre-loading
 	static {
-		testPropertiesPath = XTFConfig.getProjectRoot().resolve("test.properties");
-		globalPropertiesPath = XTFConfig.getProjectRoot().resolve("global-test.properties");
-
+		globalPropertiesPath = resolvePropertiesPath(System.getProperty(GLOBAL_TEST_PROPERTIES_PATH, "global-test.properties"));
+		testPropertiesPath = resolvePropertiesPath(System.getProperty(TEST_PROPERTIES_PATH, "test.properties"));
 		properties.putAll(XTFConfig.getPropertiesFromPath(globalPropertiesPath));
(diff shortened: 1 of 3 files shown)


core/src/main/java/cz/xtf/core/config/XTFConfig.java
```diff
@@ -29,11 +29,21 @@
 
 	private static final Properties properties = new Properties();
 
+	private static final String XTF_TEST_PROPERTIES_CONFIG_PROPERTY_NAME = "config.xtf.test.properties.path";
```
Crumby: `XTF_TEST_PROPERTIES_CONFIG_PROPERTY_NAME` -> `TEST_PROPERTIES_PATH`. Same for global test properties. `XTF` and `CONFIG` are redundant in here. `PROPERTY_NAME` by itself doesn't say anything. Also it will be consistent will all other config classes.

`config.xtf.test.properties.path` -> `xtf.test_properties.path`. Similar reasons. `config` is redundant in here and not used anywhere else. Same way all xtf properties start with `xtf`. `test_properties` rather then `test.properties` so it's clear on first sight that this is property related to this particular file. Same for global test properties.

core/src/main/java/cz/xtf/core/config/XTFConfig.java
```diff
@@ -29,11 +29,21 @@
...
+	private static final String XTF_GLOBAL_TEST_PROPERTIES_CONFIG_PROPERTY_NAME = "config.xtf.global.test.properties.path";
+
 	// Pre-loading
 	static {
-		testPropertiesPath = XTFConfig.getProjectRoot().resolve("test.properties");
-		globalPropertiesPath = XTFConfig.getProjectRoot().resolve("global-test.properties");
-
+		if (!System.getProperty(XTF_GLOBAL_TEST_PROPERTIES_CONFIG_PROPERTY_NAME,"").isEmpty()) {
```
Crumby: I'd say it will be way clearer to use:
`resolvePropertiesPath(System.getProperty(GLOBAL_TEST_PROPERTIES_PATH, "global-test.properties"))`

It's one liner with same functionality and for both (set path and default one) will be used same mechanics for resolution.

core/src/main/java/cz/xtf/core/config/XTFConfig.java
spriadka: I was worried about that `xtf.test.properties.path` would conflict with the `xtf.test.*` namespace, `xtf.test_properties.path` didn't occur to me

spriadka:
@Crumby, I fixed the code according to your suggestions. Can you take another look?

PR #276, opened 2019-05-16
record events at the end of the testsuite

maschmid:


mchoma:
+100. This will be super useful for troubelshooting. Thank you very much.

PR #281, opened 2019-06-07
Log original exception when deployment fails

mnovak1:
In tests sometimes happens that deploy fails. This will provide additional information why it happened. 

For example in:
https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OpenShift/job/eap-7.x-openshift-matrix-tests-72-matrix/47/CLUSTER_DOMAIN=multinode-nfs-005.dynamic.xpaas,label_exp=openshift-dynamic/testReport/com.redhat.xpaas.eap.hawkular/HawkularSecureConnectionTest/com_redhat_xpaas_eap_hawkular_HawkularSecureConnectionTest/

there was thrown exception after ~2 minutes but there was set 10 minutes timeout. I'm suspecting that there might be different reasons why deploy fails than just ```...Timeout while waiting for deployment of...```. If you want I can change message in exception as well to something like:
```Exception was thrown while waiting for deployment ...```


mnovak1:
@maschmid @dsimansk could you take a look at this one, please?

PR #286, opened 2019-06-11
Add image stream waiters

mchoma:
In events I can see 26x `Error resolving ImageStreamTag jboss-eap72-openjdk11-openshift-rhel8:1.0 in namespace eap72-openjdk11-tests-suite-matrix`

```
2019-06-05T16:58:16Z	2019-06-05T16:49:52Z	26	managed-executor-service.15a55b7245955036	Event	null	Warning	BuildConfigInstantiateFailed	buildconfig-controller	error instantiating Build from BuildConfig eap72-openjdk11-tests-suite-matrix/managed-executor-service (0): Error resolving ImageStreamTag jboss-eap72-openjdk11-openshift-rhel8:1.0 in namespace eap72-openjdk11-tests-suite-matrix: unable to find latest tagged image
```

This is common error in testsuite events ~40x, but in most cases event occurs 7x or 8x. In this case event occur 26x and 10 minute timeout for build was not enough.

This pull request is attempt to prepare mechanism for wait for image streams presence. Does it make sense to you?

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -1,5 +1,24 @@
 package cz.xtf.core.openshift;
 
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


maschmid:
I don't think checking imageStreamTag being not-null will be enough (I'd expect the ImageStreamTag to exist right away... it is just that its image.dockerImageReference or other fields aren't populated yet)

PR #288, opened 2019-06-20
Log what is not cleaned yet

mchoma:
We are experiencing problem that sometimes (1:100) something is not cleaned and testsuite fails on waiting. 

This is attempt to find out what is beeing stucked. So we can proceed with next steps. 

On xtf13-SNAPSHOT this is happening each time. Every time some *-deploy pod is stucked in Terminating state.

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
@@ -14,7 +14,9 @@ import cz.xtf.core.waiting.SupplierWaiter;
 import cz.xtf.core.waiting.Waiter;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.openshift.api.model.Build;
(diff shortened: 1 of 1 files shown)


Crumby:
Wait for https://github.com/xtf-cz/xtf/pull/289 and https://github.com/xtf-cz/xtf/pull/290 implementations before merging this.

mchoma:
@Crumby do you mean this should be reimplemented with  #290  capabilities?

Crumby:
> @Crumby do you mean this should be reimplemented with #290 capabilities?

Yes. Once the feature is finished. The best would be if the Info would be printed `onFailure`/`onTimeout` condition.

Crumby:
https://github.com/xtf-cz/xtf/pull/290 is now included in the code base.

mchoma:
Closing in favor of https://github.com/xtf-cz/xtf/pull/294, which seems does sufficient logging.

PR #289, opened 2019-06-21
Remove finalizers for stuck deploy pods

Crumby:


core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -966,30 +967,42 @@ public PodShell podShell(Pod pod) {
...
+		removables.addAll(configMaps().withoutLabel(OpenShift.KEEP_LABEL).list().getItems());
+		removables.addAll(getUserSecrets());
+		removables.addAll(getUserServiceAccounts());
+		removables.addAll(getUserRoleBindings());
+		removables.addAll(rbac().roles().withoutLabel(KEEP_LABEL).list().getItems());
+
+		return removables;
+	}
+
+	public void cleanFinalizers() {
```
maschmid: no need not make hacks public

core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
```diff
@@ -61,29 +63,18 @@ public Waiter isLatestBuildPresent(String buildConfigName) {
...
-			cleanedResources.add(openShift.getUserSecrets().isEmpty());
-			cleanedResources.add(openShift.getUserServiceAccounts().isEmpty());
-			cleanedResources.add(openShift.getUserRoleBindings().isEmpty());
-			cleanedResources.add(openShift.getRoles().isEmpty());
-
-			return !cleanedResources.contains(false);
+			List<HasMetadata> removables = openShift.listRemovableResources();
+
+			// HACK - Handle issue with hanging -deploy orphan
+			if (removables.isEmpty()) {
```
maschmid: isProjectClean() waiter should not contain the cleanFinalizers() hack, only the waiter returned from OpenShift::clean() should have the hack.

PR #290, opened 2019-06-21
[core] Reactive waiters

Crumby:
Extends `Waiters` with reactive API that enables custom code hookup upon different situations that may come up during waiting (success, failure, timeout, customCondition).

Examples:
- `onFailure()` can print log of the pod to show the reason of unsuccessful start
- `reactOn()` will make handling issues like https://github.com/xtf-cz/xtf/pull/289 nicer

### DIFF
--- a/core/src/main/java/cz/xtf/core/waiting/SupplierWaiter.java
+++ b/core/src/main/java/cz/xtf/core/waiting/SupplierWaiter.java
@@ -11,6 +11,11 @@ public class SupplierWaiter<X> implements Waiter {
 	private Function<X, Boolean> successCondition;
 	private Function<X, Boolean> failureCondition;
 
+	private Runnable onIteration;
+	private Runnable onSuccess;
+	private Runnable onFailure;
+	private Runnable onTimeout;
+
 	private long timeout;
 	private long interval;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/cz/xtf/core/waiting/Waiter.java
```diff
@@ -67,6 +67,30 @@
 	 */
 	Waiter logPoint(LogPoint logPoint);
 
+	/**
+	 * Sets waiters successful awaiting execution.
+	 *
+	 * @param runnable code to be executed upon successful waiting.
+	 * @return this
+	 */
+	Waiter onSuccess(Runnable runnable);
```
maschmid: adding new methods to interfaces is an API breaking change, can we at least make these _default_ ?  (alternatively, we could introduce a new interface extending the original Waiter )

core/src/main/java/cz/xtf/core/waiting/SupplierWaiter.java
```diff
@@ -86,6 +99,27 @@ public SupplierWaiter logPoint(LogPoint logPoint) {
...
+		onFailure = runnable;
+		return this;
+	}
+
+	public SupplierWaiter onTimeout(Runnable runnable) {
+		onTimeout = runnable;
+		return this;
+	}
+
+	public SupplierWaiter reactOn(Function<X, Boolean> function, Runnable runnable) {
```
maschmid: If we are not exposing the reactOn in the interface anyway (and the use case is really only for the implementation of a specific waiter, I think we can solve this with inheritance instead?

(e.g. have 

```
protected onIteration() {
  // empty
}
```

called inside waitFor on each iteration, that the speciifc waiter could override?

maschmid: (and  if you insist on having this dynamically configirable, I'd prefer to have onIteration(Runnable) ... having a specific "add if (condifition) then (runnable)" looks oddly specific.

core/src/main/java/cz/xtf/core/waiting/Waiter.java
maschmid: (or maybe not even add this into the interface at all, and keep then only in the implementation of NotSoSimpleWaiterAnymore  for now... )

Issue #293, opened 2019-07-03
OpenShift.clean() does not clean custom resources


mchoma:
I mean we are testing operator which introduce resource api `WildFlyServer.wildfly.org`. Of course xtf has no idea of such resource. 

Now what are options to resolve this? Provide method, such caller can specify additional resources he wants to remove?

Or take another approach similar to  `oc --loglevel=10 delete all --all`. Which seems first take available resource apis and then for each tries to remove resources. (but I have no idea how that can be implemented)

Crumby:
It's possible to work with `CustomResourceDefinition` without knowing its type and thus delete it. But I'm not sure that will work for all `CustomResources`.

mchoma:
Could you reference some example of working with `CustomResourceDefinition` ? Anyway it is cluster scoped resource. I would expect `OpenShift.clean()` cleans only namespace scoped resources.

dsimansk:
Some examples can be found in `fabric8io/kubernetes-client` repository.

Working with CustomResourceDefinitions
https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/CRDLoadExample.java
https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/ListCustomResourceDefinitions.java

The first approach with POJOs:
https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/CRDExample.java

The newer API levering raw approach without POJOs:
https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/RawCustomResourceExample.java

AFAIK, there's no definitive and good answer from client devs. They're working on simplifying the whole process. Anyway the former approach required to define POJOs for `CustomResources` to allow serialization/deserializationa and CRUD operations. There's no way for java client to have a model of custom resource available and the way it's implemented it can manipulate (CRUD) on such resources either.

In [Kafka Operator](https://github.com/strimzi/strimzi-kafka-operator/tree/master/api/src/main/java/io/strimzi/api/kafka/model) (Java based operator) they do generate class model the same way as in `kubernetes-client`, because they also leverage the client directly.

IMO the raw approach seems a bit more promising and easier to impl/maintain, but I haven't tried it yet.

[...]

PR #294, opened 2019-07-04
Improve OpenShift::clean reliability

dsimansk:
Related #293

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -971,7 +971,7 @@ public class OpenShift extends DefaultOpenShiftClient {
 			log.debug("DELETE :: " + hasMetadata.getKind() + "/" + hasMetadata.getMetadata().getName());
 			resource(hasMetadata).withGracePeriod(0).cascading(true).delete();
 		}
-		return new SimpleWaiter(() -> listRemovableResources().isEmpty(), TimeUnit.SECONDS, 20, "Cleaning project.");
+		return waiters.isProjectClean();
 	}
 
 	List<HasMetadata> listRemovableResources() {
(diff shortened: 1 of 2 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -967,24 +967,11 @@ public PodShell podShell(Pod pod) {
...
-
-			return false;
-		};
-
-		return new SimpleWaiter(bs, TimeUnit.SECONDS, 20, "Cleaning project.");
+		for (HasMetadata hasMetadata : listRemovableResources()) {
+			log.debug("DELETE :: " + hasMetadata.getKind() + "/" + hasMetadata.getMetadata().getName());
+			resource(hasMetadata).withGracePeriod(0).cascading(true).delete();
+		}
+		return new SimpleWaiter(() -> listRemovableResources().isEmpty(), TimeUnit.SECONDS, 20, "Cleaning project.");
```
Crumby: I think we can switch back to default one present in [OpenShiftWaiters](https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java#L60)

dsimansk: Done

mchoma:
@dsimansk was there any related fix in 4.3.0 client? Can you reference? I dont see how does it relate to #293, we still dont have custom resources in `listRemovableResources` and I dont assume any of standard recources will cascade to delete CR. What am I missing?

maschmid:
I suppose he meant to link to the issue describing the problem with hang -deploy pods on namespace cleanup...

mchoma:
for record: ```fabric8 client sends  {"apiVersion":"v1","kind":"DeleteOptions","gracePeriodSeconds":0,"orphanDependents":true}   when cascading is false
see also https://github.com/fabric8io/kubernetes-client/issues/1614```

dsimansk:
@mchoma Marek is correct. I've mixed the issues together sorry. This PR is aiming purely on stuck resources with orphan finalizers.

PR #295, opened 2019-07-09
Build Config timeout is configured independently. Unify to 2 minutes.

mchoma:
Feature was introduced by https://github.com/fabric8io/kubernetes-client/pull/501.

Happened to us building pod was evicted because of DiskPresure for 15 sec. Default timeout seems to be 10 seconds. This change can help solve this cases.

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -74,6 +74,8 @@ public class OpenShift extends DefaultOpenShiftClient {
 
 	public static final String KEEP_LABEL = "xtf.cz/keep";
 
+	public static final int DEFAULT_TIMEOUT = 120_000;
+
 	/**
 	 * Autoconfigures the client with the default fabric8 client rules
 	 * @return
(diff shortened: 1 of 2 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -74,6 +74,8 @@
 
 	public static final String KEEP_LABEL = "xtf.cz/keep";
 
+	public static final int DEFAULT_TIMEOUT = 120_000;
```
maschmid: If we are going to make it public, we should probably have a more descriptive name and some javadoc for it.  Perhaps DEFAULT_HTTP_TIMEOUT ?  This value is basically used for a mix of various timeouts related to okhttp (connect, read and write timeouts) and the only reason they are being set is binary s2i builds.

mchoma:
I think the idea was to align BuildConfig timeout with OpenShiftConfig

```
config.setRequestTimeout(120_000);
config.setConnectionTimeout(120_000);
```
If I will make timeout constant private in BuildConfig I cant use it in OpenShiftConfig.

But we do not experience this problem anymore, so I am going to close this.

Issue #296, opened 2019-07-16
Problem when getting logs of a pod with many containers - Suggestion for API improvement


radtriste:
Trying to access the logs from a pod with many containers, I got this error:

> io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://##SERVER##/api/v1/namespaces/##NAMESPACE##/pods/##POD_NAME##/log?pretty=false. Message: a container name must be specified for pod ##POD_NAME##, choose one of: [##CONTAINER1##, ##CONTAINER2##]. Received status: Status(apiVersion=v1, code=400, details=null, kind=Status, message=a container name must be specified for pod ##POD_NAME##, choose one of: [##CONTAINER1##, ##CONTAINER2##], metadata=ListMeta(_continue=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=BadRequest, status=Failure, additionalProperties={}).
at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:478)
at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:417)
at io.fabric8.kubernetes.client.dsl.internal.PodOperationsImpl.doGetLog(PodOperationsImpl.java:150)
at io.fabric8.kubernetes.client.dsl.internal.PodOperationsImpl.getLog(PodOperationsImpl.java:159)
at io.fabric8.kubernetes.client.dsl.internal.PodOperationsImpl.getLog(PodOperationsImpl.java:61)
	at cz.xtf.core.openshift.OpenShift.getPodLog(OpenShift.java:285)

The API does not provide yet a way to get logs from a specific container, or all the container logs from a pod.

The workaround is to use the inner code of `getPodLog` method and do:
`openshift.pods().withName(name).inContainer(##CONTAINER_NAME##).getLog()` to retrieve exact logs of a container.

API improvements (Backward compatible):
- `getPodLog` => Return the logs from first container ? Would it be a problem ?
- `getPodAllContainersLog` => Return all logs as a collection of String
- `getPodContainerLog(String name, String containerName)` => Return specific log of a specific container

Container names can anyway be retrieved from the Pod class (via `pod.getSpec().getContainers()` and `container.getName()`).

_NB: This does also apply for `observePodLog` methods._

If needed, I can provide a PR for it. Just let me know.

PR #304, opened 2019-09-05
[core] Added support for multiple containers in a pod for logs

radtriste:
Issue #296: Problem when getting logs of a pod with many containers -
Suggestion for API improvement

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -301,7 +301,7 @@ public class OpenShift extends DefaultOpenShiftClient {
 	}
 	
 	public String getPodLog(Pod pod, String containerName) {
-		return getPodLog(pod, getContainerByName(pod, containerName));
+		return getPodLog(pod, getContainer(pod, containerName));
 	}
 
 	public String getPodLog(Pod pod, Container container) {
@@ -319,11 +319,15 @@ public class OpenShift extends DefaultOpenShiftClient {
 	 * @return Map of container name / logs
 	 */
 	public Map<String, String> getAllContainerLogs(Pod pod) {
-		return retrieveFromPodContainer(pod, container -> this.getPodLog(pod, container));
+		return retrieveFromPodContainers(pod, container -> this.getPodLog(pod, container));
 	}
 
 	public Reader getPodLogReader(Pod pod) {
 		return getPodLogReader(pod, getAnyContainer(pod));
+	}
+	
+	public Reader getPodLogReader(Pod pod, String containerName) {
+		return getPodLogReader(pod, getContainer(pod, containerName));
 	}
 
 	public Reader getPodLogReader(Pod pod, Container container) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -372,6 +444,38 @@ public boolean deletePods(String key, String value) {
...
+	}
+	
+	public Container getContainerByName(Pod pod, String containerName) {
+		return getAllContainers(pod).stream()
+			.filter(c -> c.getName().equals(containerName))
+			.findFirst()
+			.orElseThrow(() -> new RuntimeException("Cannot find container with name " + containerName + " in pod " + pod.getMetadata().getName()));
+	}
+
+	private <R> Map<String, R> retrieveFromPodContainer(Pod pod, Function<Container, R> containerRetriever) {
```
Crumby: Please make a method name plural so it's consistent with return type (`retrieveFromPodContainer`). This way it seems it should be from one container.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -372,6 +444,38 @@ public boolean deletePods(String key, String value) {
...
+	 * 
+	 * @param pod Pod to retrieve from
+	 * @return One random container from the pod
+	 */
+	public Container getAnyContainer(Pod pod) {
+		List<Container> containers = getAllContainers(pod);
+		return containers.get(new Random().nextInt(containers.size()));
+	}
+	
+	public Container getContainerByName(Pod pod, String containerName) {
```
Crumby: Make it just `getContainer` so it's consistent with other getResource methods.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -292,21 +297,88 @@ public String getPodLog(String deploymentConfigName) {
...
+	public Map<String, String> getAllContainerLogs(Pod pod) {
+		return retrieveFromPodContainer(pod, container -> this.getPodLog(pod, container));
 	}
 
 	public Reader getPodLogReader(Pod pod) {
-		return pods().withName(pod.getMetadata().getName()).getLogReader();
+		return getPodLogReader(pod, getAnyContainer(pod));
+	}
+
+	public Reader getPodLogReader(Pod pod, Container container) {
```
Crumby: Please reorganize the method so the final its top down from the one with the simplest parameters to the actual implementation:
```
getPodLogReader(Pod pod)
getPodLogReader(Pod pod, String containerName)
getPodLogReader(Pod pod, Container container)
```

Applies also for others updated and created methods.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
radtriste: updated

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
radtriste: updated

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
radtriste: updated

Issue #305, opened 2019-09-05
[bug] Openshift waiter "hasBuildCompleted" with buildConfigName is failing if build is not yet existing ...


radtriste:
I came to the case where I create the build via the command line "oc start-build ..." and then try to wait for the build to be completed, using this method in OpenshiftWaiter class:

```java
public Waiter hasBuildCompleted(String buildConfigName) {
	return hasBuildCompleted(openShift.getLatestBuild(buildConfigName));
}
```

This is calling:

```java
public Waiter hasBuildCompleted(Build build) {
	Supplier<String> supplier = () -> openShift.getBuild(build.getMetadata().getName()).getStatus().getPhase();
	String reason = "Waiting for completion of build " + build.getMetadata().getName();

	return new SupplierWaiter<>(supplier, "Complete"::equals, "Failed"::equals, TimeUnit.MINUTES, 10, reason).logPoint(Waiter.LogPoint.BOTH).interval(5_000);
}
```

There I can a `NullPointerException` because there is no latest build existing and thus, the build sent to this method is null.

I say we shall wait for the latest build to exist (or check it is the corect latest build ...)

mchoma:
Can you use `OpenShiftWaiters.isLatestBuildPresent()` ?

radtriste:
This one is indeed useful. I missed it ...

But still, from `OpenShiftWaiters.hasBuildCompleted(String buildConfigName)`, I would expect it is waiting that the build is created before getting it and waiting for it...

radtriste:
I can make a quick PR for that if needed.

PR #311, opened 2019-09-06
[bug] Small correction on waitForBuildCompleted if buildconfig not yet existing 

radtriste:
issue #305

Wait for latest build to exist

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
@@ -2,6 +2,7 @@ package cz.xtf.core.openshift;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
```diff
@@ -21,6 +21,7 @@
 	}
 
 	public Waiter hasBuildCompleted(String buildConfigName) {
+		isLatestBuildPresent(buildConfigName).waitFor();
```
Crumby: Actual waiting goes against convention of this class. None of the methods waits for anything, they only return executable Waiter objects.

radtriste: ok.. Need to set the documentation then.

radtriste: I will update this

Crumby: Updating the docs is good but you can also fix the method so the user doesn't have to call `isLatestBuildPresent` and then this one by rewriting the Waiter implementation for this method. It'll be combination of the one isLatestBuildPresent and the one that it calls right now.

radtriste: ok. Did a simple implementation using Optional and returning null string if not yet existing.

Could also creates a SimpleWaiter with the condition with condition not null and Phase Complete or Failed but thought it clearer and in sync with the other 'hasBuildCompleted' method. Let me know what you think of it.

radtriste:
squash and rebase done.

PR #314, opened 2019-09-24
Revert "[core] refactor OpenShift::clean to use Background policyDele…

dsimansk:
…gation"

This reverts commit caab6c69c04583a9c24b021aa5b02bc9266cdf44.

The newly added method for propagation policy is not setting any property on delete req. I'll open new issue in kubernetes-client repo.

Payload is still the same (controlled only by `cascading()`):
```
INFO: {"apiVersion":"v1","kind":"DeleteOptions","gracePeriodSeconds":0,"orphanDependents":false}
```
or
```
INFO: {"apiVersion":"v1","kind":"DeleteOptions","gracePeriodSeconds":0,"orphanDependents":true}
```

Cc @mchoma


dsimansk:
I've left the client upgrade commit untouched intentionally.

Issue #312, opened 2019-09-11
Relax CRD clean in case CRD is not installed


pkremens:
https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShift.java#L979
```
	public Waiter clean() {
	for (CustomResourceDefinitionContextProvider crdContextProvider : OpenShift.getCRDContextProviders()) {
			customResource(crdContextProvider.getContext()).delete(getNamespace());
	}
```
We install the operator CRD during the test execution, however the cleanup blows in case that CRD in not installed on cluster yet (e.g. first run on a clean cluster). We'd need relax this part a little bit so it doesn't fail in case that CRD is not installed yet.

pkremens:
Or at least to have some other way to control this - sorry for vague description, I'm in pre-PTO time press, I'll add more details once I'm back.

pkremens:
I'm getting back on this one. This is what I get at the moment (0.14-20190924.141535-5)
```
io.fabric8.kubernetes.client.KubernetesClientException: Error while performing the call to https://192.168.130.11:6443/apis/wildfly.org/v1alpha1/namespaces/pkremens-namespace/wildflyservers/. Response code: 403
	at io.fabric8.kubernetes.client.dsl.internal.RawCustomResourceOperationsImpl.makeCall(RawCustomResourceOperationsImpl.java:230)
	at io.fabric8.kubernetes.client.dsl.internal.RawCustomResourceOperationsImpl.delete(RawCustomResourceOperationsImpl.java:154)
	at cz.xtf.core.openshift.OpenShift.clean(OpenShift.java:985)
```

Crumby:
Yeah, the `clean()` actually expects that CRD is already installed on the cluster before running the tests. But we can't ruin anything putting it into try/catch block. Will you create a PR with the fix?

pkremens:
Sure, I'll look into it and let you know once I have something (ultimately I'll fall back to try/catch in case I won't be able to find some more intelligent solution).

pkremens:
I also get this one in case that CRD is installed, but no instance of it is running.
```
io.fabric8.kubernetes.client.KubernetesClientException: Error while performing the call to https://192.168.130.11:6443/apis/wildfly.org/v1alpha1/namespaces/pkremens-namespace/wildflyservers/. Response code: 403
[...]

PR #315, opened 2019-09-26
[core] [#312] Relax CRD clean in case CRD is not installed

pkremens:
https://github.com/xtf-cz/xtf/issues/312

I run into several problems with this one

1) KubernetesClientException in case that CRD is not registered
2) KubernetesClientException in case that no CRD instances exists - this one could happen if CRD resources are not cleaned properly (e.g. due to finalizers), but the role-bindings and other service accounts are, at this point listing the resources throws KubernetesClientException
3) `customResource(crdContextProvider.getContext()).delete(getNamespace());` doesn't seem to block anymore (I would bet that it was before, but we're too close to freeze now, so I cannot investigate more now), thus we need to wait after deleting, otherwise TS could run into issues like:
```
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: 
https://192.168.130.11:6443/apis/wildfly.org/v1alpha1/namespaces/pkremens-namespace
/wildflyservers. Message: object is being deleted: wildflyservers.wildfly.org "wildfly" already exists. 
```

@Crumby @dsimansk guys please take a look and let me know wdyt, although I myself do not like the `OpenShifts.admin()` part much, I do not know a better way to do this.

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -151,7 +151,7 @@ public class OpenShift extends DefaultOpenShiftClient {
 		config.setConnectionTimeout(120_000);
 	}
 
-	private static synchronized ServiceLoader<CustomResourceDefinitionContextProvider> getCRDContextProviders() {
+	protected static synchronized ServiceLoader<CustomResourceDefinitionContextProvider> getCRDContextProviders() {
 		if (crdContextProviderLoader == null) {
 			crdContextProviderLoader = ServiceLoader.load(CustomResourceDefinitionContextProvider.class);
 		}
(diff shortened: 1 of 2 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -982,7 +993,12 @@ public PodShell podShell(Pod pod) {
 	 */
 	public Waiter clean() {
 		for (CustomResourceDefinitionContextProvider crdContextProvider : OpenShift.getCRDContextProviders()) {
-			customResource(crdContextProvider.getContext()).delete(getNamespace());
+			CustomResourceDefinitionContext crdContext = crdContextProvider.getContext();
+			if (getCRDItems(crdContext, getNamespace()).size() > 0) {
+				log.debug("DELETE :: " + crdContext.getName() + " instances");
+				customResource(crdContext).delete(getNamespace());
+				new SimpleWaiter(() -> getCRDItems(crdContext, getNamespace()).size() == 0, TimeUnit.SECONDS, 20, "Cleaning CRDs.").waitFor();
```
pkremens: This might not be enough (20 seconds) as Operator can create a lot of objects and pod which took time to terminate. I'm testing this atm.

pkremens: Ok, 20s doesn't seem to be a problem here, my problem was a finalizer...

Crumby: No waiting should be executed within the function. But you are right that there is a bug and `isProjectClean()` doesn't take custom resources into account. `isProjectClean()` should be fixed for this purpose. For the finalizer, is there possibility to do the same as with regular resources (cc @dsimansk )?

pkremens:
@Crumby How about this one https://github.com/xtf-cz/xtf/pull/315/commits/f3f999d48218999b3a66fbf23d617adbb2e68e48 ?

pkremens:
@Crumby done.

PR #318, opened 2019-10-04
Only test given path in ImageContent#listZipFilesInDir

jbliznak:
It seems that is a bug to also test `~` (user home) along with given path.
One should be able to only test the desired path.


jbliznak:
Ok I lost this race, duplicate of https://github.com/xtf-cz/xtf/pull/317

Issue #308, opened 2019-09-06
0.14-SNAPSHOT is generating WaiterException: CleaningProject


mchoma:
We are missing way to specify deletion policy in kubernetes client  [1]. 
(And probably with 0.13 default becomes orphan policy??)

Anyway, currently we are experiencing `*-deploy` pods not beeing deleted because they have `finalizer: orphan` and `ownerReference` filled to parent object (I think it was ReplicaSet) in same time.

So either
1.  `OpenShift.clean()` does not clean parent objects properly. Should we return to `cascading(true)` ?
2. Or there is some bug in kubernetes GC and `ownerReference` is not cleared properly from child objects

Shouldnt we apply `cleanFinalizers` workaround [2] again to cover this scenario?

[1] https://github.com/fabric8io/kubernetes-client/issues/1614
[2] https://github.com/xtf-cz/xtf/commit/caa00a4d3d9b28adf6514da80046c9dad102675a

Crumby:
Are you using same fabric8 client version as the xtf release? `-deploy` orphans shouldn't be a problem with 4.3.1.

mchoma:
With aligning client versions this problem is gone. What are we experiencing now is much more longer testsuite run - it has nearly doubled. From nearly 5 hours to 9 hours. Have you noticed something similar?

mchoma:
@dsimansk Now with 0.14-SNAPSHOT I run into situation that ReplicaSet was not deleted `Replica Sets jaeger-7c76d4c55`, thus jaeger pod keep restarting. 

```
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  annotations:
    deployment.kubernetes.io/desired-replicas: '1'
    deployment.kubernetes.io/max-replicas: '1'
    deployment.kubernetes.io/revision: '1'
  creationTimestamp: '2019-10-03T07:43:01Z'
  generation: 1
  labels:
    app: jaeger
    jaeger-infra: jaeger-pod
    pod-template-hash: '373280711'
  name: jaeger-7c76d4c55
  namespace: eap72-tests-13-mx
  resourceVersion: '304197'
  selfLink: /apis/apps/v1/namespaces/eap72-tests-13-mx/replicasets/jaeger-7c76d4c55
  uid: 6d4a5667-e5b1-11e9-bd82-fa163e733bcb
```

After manual delete through web console problem was away.

Crumby:
[...]

PR #319, opened 2019-10-11
[#308] Clean does not remove ReplicaSets

mchoma:
https://github.com/xtf-cz/xtf/issues/308#issuecomment-537833455

What I am not sure is order of cleaning ReplicaSet. Does someone see problem with putting it in front to `Deployments`, `DeploymentConfigs`, `StatefulSets`, `ReplicationControllers`? I believe this object belongs to this set.

Running localy testsuite with this change in place.


dsimansk:
I'd suggest to put it after `deployments()` listing to copy the order from `DeploymentConfig` and `ReplicationController`.

mchoma:
@dsimansk make sense.

So reasoning is Deployments creates ReplicaSets. So lets delete first Deployments and then ReplicaSets. 

Anyway, running our testsuite locally order replicasets, deployments did no problem.

PR #321, opened 2019-10-16
Refactor OpenShift:clean to delete resources via list

dsimansk:
Good old times are coming back...


mchoma:
Won't that introduce `WaiterException: Cleaning project `?

dsimansk:
I brought back "safety net" code. From my limited test executions the performance degradation is gone and I didn't see leftovers warning either. 
Performance issue is definitely caused by different handling of list vs. single resource deletion. I'll try to collect more data tomorrow.

dsimansk:
@Crumby would you mind merging this PR and trying it out in the wild? 

After a brief discussion with @maschmid, the fallback plan will be to implement `clean()` through plain http calls with the same `DeleteOptions` etc. as `oc delete all --all`.

mchoma:
@dsimansk  Could we have release with that change in place? Running our testsuite we do not see `WaiterException: Cleaning project` anymore and performance is ~ same.

PR #322, opened 2019-10-18
Add OpenShift::setupPullSecret + config

### LINKED ISSUE #283: Allow testsuite to pull images from secured repo e.g. `registry.redhat.io`
This seemed to be in https://github.com/xtf-cz/xtf/pull/153/files however was removed by https://github.com/xtf-cz/xtf/commit/dafcd8def33d039256e88e431befac4acb412fa5 .

dsimansk:
Resolves #283 

This PR adds support for pull secret in json format and mimics the openshift-install `pullSecret` parameter. It's quiet extensible to add additional registries e.g. quay, dockerhub there's no need to change the impl.

Please review and let me know if that's sufficient solution. I've tested it with in namespace deployment and with `BuildManager` successfully.

Format:
```json
{"auths":{"registry.redhat.io":{"auth":"<REDACTED_TOKEN>"}}}
```

### DIFF
--- a/core/src/main/java/cz/xtf/core/bm/BuildManager.java
+++ b/core/src/main/java/cz/xtf/core/bm/BuildManager.java
@@ -1,6 +1,7 @@
 package cz.xtf.core.bm;
 
 import cz.xtf.core.config.BuildManagerConfig;
+import cz.xtf.core.config.OpenShiftConfig;
 import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.waiting.Waiter;
 import lombok.extern.slf4j.Slf4j;
@@ -14,6 +15,10 @@ public class BuildManager {
 
 		if (openShift.getProject(openShift.getNamespace()) == null) {
 			openShift.createProjectRequest();
+			openShift.waiters().isProjectReady().waitFor();
+		}
+		if (OpenShiftConfig.pullSecret() != null) {
+			openShift.setupPullSecret(OpenShiftConfig.pullSecret());
 		}
 
 		openShift.addRoleToGroup("system:image-puller", "ClusterRole",  "system:authenticated");
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -13,14 +13,17 @@ import io.fabric8.kubernetes.api.model.Event;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.HorizontalPodAutoscaler;
 import io.fabric8.kubernetes.api.model.KubernetesList;
+import io.fabric8.kubernetes.api.model.LocalObjectReferenceBuilder;
 import io.fabric8.kubernetes.api.model.Node;
 import io.fabric8.kubernetes.api.model.ObjectMetaBuilder;
+import io.fabric8.kubernetes.api.model.ObjectReferenceBuilder;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodSpec;
 import io.fabric8.kubernetes.api.model.ReplicationController;
 import io.fabric8.kubernetes.api.model.ResourceQuota;
 import io.fabric8.kubernetes.api.model.Secret;
+import io.fabric8.kubernetes.api.model.SecretBuilder;
 import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.api.model.ServiceAccount;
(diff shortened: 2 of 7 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -166,6 +169,25 @@ public OpenShift(OpenShiftConfig openShiftConfig) {
...
+		Secret pullSecret = new SecretBuilder()
+				.withNewMetadata()
+					.withNewName(name)
+					.addToLabels(KEEP_LABEL, "true")
+				.endMetadata()
+				.withNewType("kubernetes.io/dockerconfigjson")
+				.withData(Collections.singletonMap(".dockerconfigjson", Base64.getEncoder().encodeToString(secret.getBytes())))
+				.build();
+		secrets().createOrReplace(pullSecret);
+		serviceAccounts().withName("default").edit().addToSecrets(new ObjectReferenceBuilder().withName(pullSecret.getMetadata().getName()).build()).done();
```
maschmid: according to https://docs.openshift.com/container-platform/4.2/openshift_images/managing-images/using-image-pull-secrets.html

we should 

* add it to the "secrets" list of the builder SA
* add it to the "imagePullSecrets" list of the default SA 

oc secrets link default <pull_secret_name> --for=pull
oc secrets link builder <pull_secret_name>

creates something like this:

```
$ oc get sa default -o yaml
apiVersion: v1
imagePullSecrets:
- name: default-dockercfg-cqfww
- name: mysecret
kind: ServiceAccount
metadata:
  creationTimestamp: "2019-10-16T11:00:03Z"
  name: default
  namespace: myproject
  resourceVersion: "1794373"
  selfLink: /api/v1/namespaces/myproject/serviceaccounts/default
  uid: 1ac74c1c-f004-11e9-a321-06b56ce0f943
secrets:
- name: default-dockercfg-cqfww
- name: default-token-tcf9j



[maschmid@archie ~]$ oc get sa builder -o yaml
apiVersion: v1
imagePullSecrets:
- name: builder-dockercfg-88ss6
kind: ServiceAccount
metadata:
  creationTimestamp: "2019-10-16T11:00:03Z"
  name: builder
  namespace: myproject
  resourceVersion: "1794457"
  selfLink: /api/v1/namespaces/myproject/serviceaccounts/builder
  uid: 1ac7896e-f004-11e9-8d2d-0603d46595f1
secrets:
- name: builder-token-dlvqs
- name: builder-dockercfg-88ss6
- name: mysecret

```

dsimansk: Ok, will do. Keep or remove from `deployer` SA?

dsimansk: Done

core/src/main/java/cz/xtf/core/bm/BuildManager.java
```diff
@@ -15,6 +16,9 @@ public BuildManager(OpenShift openShift) {
 		if (openShift.getProject(openShift.getNamespace()) == null) {
 			openShift.createProjectRequest();
 		}
+		if (OpenShiftConfig.pullSecret() != null) {
```
maschmid: I suspect there may be a race between the project being created and the creation of the secret in the namespace.   ( openShift.createProjectRequest(); is async afaik)

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
```diff
@@ -19,6 +20,9 @@ public void testPlanExecutionStarted(TestPlan testPlan) {
 		if(openShift.getProject() == null) {
 			openShift.createProjectRequest();
 		}
+		if(OpenShiftConfig.pullSecret() != null) {
```
maschmid: also here, 

I suspect there may be a race between the project being created and the creation of the secret in the namespace. ( openShift.createProjectRequest(); is async afaik, so I suspect the creation may fail if the namespace isn't created by then,)

core/src/main/java/cz/xtf/core/bm/BuildManager.java
dsimansk: Yes, that's true. I was planing to create something like `isProjectReady` waiter.

dsimansk:
@maschmid updated with `OpenShiftWaiters::isProjectReady`.

dsimansk:
Just another idea, we could have a method createProject that would return waiter same as`clean()`.

core/src/main/java/cz/xtf/core/bm/BuildManager.java
dsimansk: I'll mark this as resolved through `isProjectReady()` addition. Let me know if you don't like it. :)

mchoma:
It is ok from my point of view.

dsimansk:
@maschmid thanks :)

Issue #316, opened 2019-10-02
Add log level to Waiter interface (enhancement)


pkremens:
These would come handy:
```
new SimpleWaiter(() -> isConditionMet(), "Wait for condition to be met.", Level.DEBUG)
	.interval(100L)
	.waitFor();
```

https://stackoverflow.com/a/3831939

dsimansk:
IMO this is great idea. @pkremens would you like to implement it? :)

pkremens:
I'd really like to :), but I have no capacity to do it at the moment. I'll keep that one in my personal backlog and once I clean a table a little bit, I can look into it (no promises here). Nonetheless in case someone wants to take on it, it's free to be taken ;).

PR #326, opened 2019-11-01
Extend the Waiter interface with a level() 

pkremens:
https://github.com/xtf-cz/xtf/issues/316

```
new SimpleWaiter(() -> isConditionMet(), "Wait for condition to be met.")
	.interval(100L)
	.level(Level.DEBUG)
	.waitFor();
```

*Waiter constructors could be extended as well if desired.

### DIFF
--- a/core/src/main/java/cz/xtf/core/config/WaitingConfig.java
+++ b/core/src/main/java/cz/xtf/core/config/WaitingConfig.java
@@ -1,11 +1,19 @@
 package cz.xtf.core.config;
 
+import org.slf4j.event.Level;
+
 public class WaitingConfig {
 	public static final String WAITING_TIMEOUT = "xtf.waiting.timeout";
+	public static final String WAITING_LOG_LEVEL = "xtf.waiting.log.level";
 
 	private static final String WAITING_TIMEOUT_DEFAULT = "180000";
+	private static final String WAITING_LOG_LEVEL_DEFAULT = "INFO";
 
 	public static long timeout() {
 		return Long.parseLong(XTFConfig.get(WAITING_TIMEOUT, WAITING_TIMEOUT_DEFAULT));
+	}
+
+	public static Level level() {
+		return Level.valueOf(XTFConfig.get(WAITING_LOG_LEVEL, WAITING_LOG_LEVEL_DEFAULT).toUpperCase());
 	}
 }
--- a/core/src/main/java/cz/xtf/core/waiting/Waiter.java
+++ b/core/src/main/java/cz/xtf/core/waiting/Waiter.java
@@ -1,6 +1,8 @@
 package cz.xtf.core.waiting;
 
+import cz.xtf.core.config.WaitingConfig;
 import org.apache.commons.lang3.time.DurationFormatUtils;
+import org.slf4j.event.Level;
 
 import java.util.concurrent.TimeUnit;
 
@@ -68,6 +70,17 @@ public interface Waiter {
 	Waiter logPoint(LogPoint logPoint);
 
 	/**
+	 * Set the level of severity for a log message.
+	 *
+	 * @param level what level of severity should be used to log the message.
+	 * @return this
+	 * @see Level
+	 */
+	default Waiter level(Level level) {
+		throw new UnsupportedOperationException("Method level hasn't been implemented.");
+	}
+
+	/**
 	 * Sets waiters execution for each iteration.
 	 *
 	 * @param runnable code to be executed upon successful waiting.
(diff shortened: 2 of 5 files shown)


mchoma:
Nice one.

Crumby:
Maybe it wouldn't be bad to have global configuration property for this?

pkremens:
You mean for `DEFAULT_LEVEL`? I'm not sure that it is desired, at least not for our project.

pkremens:
Nonetheless it can be added if required...

core/src/main/java/cz/xtf/core/waiting/Waiter.java
```diff
@@ -67,6 +69,15 @@
...
 	Waiter logPoint(LogPoint logPoint);
 
+	/**
+	 * Set the level of severity for a log message.
+	 *
+	 * @param level what level of severity should be used to log the message.
+	 * @return this
+	 * @see Level
+	 */
+	Waiter level(Level level);
```
dsimansk: It might be better to do a default interface impl. To prevent any breaking changes of custom waiters. I'd say there's minimal chance of custom waiters being present in the wild, but we all know the drill... :)

@Crumby wdyt?

dsimansk:
> Maybe it wouldn't be bad to have global configuration property for this?

Agreed, a bit of future proofing just in case.

core/src/main/java/cz/xtf/core/waiting/Waiter.java
Crumby: @dsimansk Agreed. We can't break anything by making this default similarly to `onSuccess()`.

pkremens: So

```
	default Waiter level(Level level) {
		throw new UnsupportedOperationException("Method onIteration hasn't been implemented.");
	}
```

would it be?

pkremens: Fixed

pkremens:
@dsimansk @Crumby as for global property goes, is it ok to reuse the `WaitingConfig` for this (e.g. a new level() method similar to timeout())?

Crumby:
@pkremens Do you think you could you extend the implementation with the global property while you are at it (defaulting to INFO...)? It should only take adding `xtf.waiting.log.level` property and parser (`Level.valueOf(String value)`) to [WaitingConfig](https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/config/WaitingConfig.java) class and pointing `DEFAULT_LEVEL` to that property.

Doesn't change the functionality but if user would like to default debug globally and optionally set info then it would be more convenient then having to set `level(Level.DEBUG)` on each waiter.

Otherwise PR looks good to me.

pkremens:
@dsimansk take a look at https://github.com/xtf-cz/xtf/pull/326/commits/c740cc33328ceb7b37af34aa81e8cc884b895366

dsimansk:
@pkremens I like the property name `xtf.waiting.log.level` a bit more. Otherwise it's OK.

pkremens:
Updated to `xtf.waiting.log.level` and squashed.

core/src/main/java/cz/xtf/core/config/WaitingConfig.java
```diff
@@ -1,11 +1,22 @@
...
+
+	private static final Level WAITING_LOG_LEVEL_DEFAULT = Level.INFO;
+
 	public static long timeout() {
 		return Long.parseLong(XTFConfig.get(WAITING_TIMEOUT, WAITING_TIMEOUT_DEFAULT));
 	}
+
+	public static Level level() {
+		String level = XTFConfig.get(WAITING_LOG_LEVEL);
+		return (level == null) ? WAITING_LOG_LEVEL_DEFAULT : Level.valueOf(level.toUpperCase());
```
Crumby: There is method `XTFConfig.get(String property, String default)` that does the if part for you and hides the 'ugly' code. It'd be nicer to use that.

One liner would than look like:
`return Level.valueOf(XTFConfig.get(WAITING_LOG_LEVEL, WAITING_LOG_LEVEL_DEFAULT).toUpperCase() )`

pkremens: I saw that one, but I didn't want to type the WAITING_LOG_LEVEL_DEFAULT as String

core/src/main/java/cz/xtf/core/config/WaitingConfig.java
```diff
@@ -1,11 +1,22 @@
 package cz.xtf.core.config;
 
+import org.slf4j.event.Level;
+
 public class WaitingConfig {
 	public static final String WAITING_TIMEOUT = "xtf.waiting.timeout";
 
 	private static final String WAITING_TIMEOUT_DEFAULT = "180000";
 
+	public static final String WAITING_LOG_LEVEL = "xtf.waiting.log.level";
```
Crumby: Can you group the fields first by properties and then by default values? I find that a more consistent with other classes and easier to read.

core/src/main/java/cz/xtf/core/config/WaitingConfig.java
pkremens: but it may work after all, let me try...

pkremens: I have to either 
```
return Level.valueOf(XTFConfig.get(WAITING_LOG_LEVEL, WAITING_LOG_LEVEL_DEFAULT.toString()).toUpperCase());
```
or change 
```
	private static final Level WAITING_LOG_LEVEL_DEFAULT = Level.INFO;
->
	private static final String WAITING_LOG_LEVEL_DEFAULT = "INFO";

```
since WAITING_LOG_LEVEL_DEFAULT I'm fine with making it String instead

core/src/main/java/cz/xtf/core/config/WaitingConfig.java
pkremens: Good point, done

core/src/main/java/cz/xtf/core/config/WaitingConfig.java
Crumby: I'd go with `String` the same way timeout is `String`. They are properties in the end which are always strings by nature.

pkremens: Sure, no problem, it's in.

Crumby:
Merged. Thanks @pkremens!

pkremens:
@Crumby @dsimansk Thanks for revision!

PR #327, opened 2019-11-01
[EnvVars] Avoid NPE in env vars when the value is null

avano:
continuing of https://github.com/xtf-cz/xtf/pull/146 , don't use `Collectors.toMap` in `getDeploymentConfigEnvVars` as that doesn't work with `null` values

ref: https://stackoverflow.com/a/24632080


dsimansk:
Thanks, @avano!

Issue #328, opened 2020-01-07
Make clean project timeout configurable


mchoma:
Currently it is hardcoded to 20 seconds. 

There are scenarios where it is necessary to wait longer. Since Kubernetes 1.16 (OCP 4.3) by default Service of type LoadBalancer waits until cloud providers load balancer resource is cleaned up [1]. On AWS 20 seconds is sometimes not enough.

[1] https://v1-16.docs.kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#garbage-collecting-load-balancers

mchoma:
Fixed by https://github.com/xtf-cz/xtf/pull/329

PR #329, opened 2020-01-07
[#328] Make clean project timeout configurable

mchoma:
issue: https://github.com/xtf-cz/xtf/issues/328

### DIFF
--- a/core/src/main/java/cz/xtf/core/config/WaitingConfig.java
+++ b/core/src/main/java/cz/xtf/core/config/WaitingConfig.java
@@ -5,9 +5,11 @@ import org.slf4j.event.Level;
 public class WaitingConfig {
 	public static final String WAITING_TIMEOUT = "xtf.waiting.timeout";
 	public static final String WAITING_LOG_LEVEL = "xtf.waiting.log.level";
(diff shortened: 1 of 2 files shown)


core/src/main/java/cz/xtf/core/config/WaitingConfig.java
```diff
@@ -5,9 +5,11 @@
 public class WaitingConfig {
 	public static final String WAITING_TIMEOUT = "xtf.waiting.timeout";
 	public static final String WAITING_LOG_LEVEL = "xtf.waiting.log.level";
+	public static final String CLEAN_PROJECT_TIMEOUT = "xtf.clean.project.timeout";
```
Crumby: @mchoma Can you make it `xtf.waiting.timeout.cleanup` (alternatively `clean`, `clean_project`, `project_cleanup`)? Property will be more consistent with others. Also I think `cleanup` should be enough, I can't recall any other cleanup we do in the framework atm.

mchoma: I have changed to `xtf.waiting.timeout.cleanup` please have a look

PR #330, opened 2020-01-16
[fix] Wait for deletion in BinaryBuild

llowinge:


core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -107,7 +110,12 @@ public void update(OpenShift openShift) {
 	public void delete(OpenShift openShift) {
 		openShift.imageStreams().withName(is.getMetadata().getName()).cascading(false).withGracePeriod(0).delete();
 		openShift.buildConfigs().withName(bc.getMetadata().getName()).cascading(true).withGracePeriod(0).delete();
-		openShift.pods().withName(bc.getMetadata().getName() + "-1-build").cascading(false).withGracePeriod(0).delete();
+		final String podName = bc.getMetadata().getName() + "-1-build";
+		openShift.pods().withName(podName).cascading(false).withGracePeriod(0).delete();
+
+		new SimpleWaiter(() -> openShift.getImageStream(is.getMetadata().getName())  == null, TimeUnit.MILLISECONDS, WaitingConfig.timeout(), "Waiting for old imageStreams deletion before creating new one").waitFor();
```
mchoma: with your waiting message you are assuming BinaryBuild.delete() is used with BinaryBuild.create().

But this does not have to be always true. Can you use just something like `Waiting for  imageStream deletion`

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -107,7 +110,12 @@ public void update(OpenShift openShift) {
...
+		new SimpleWaiter(() -> openShift.getBuildConfig(bc.getMetadata().getName())  == null, TimeUnit.MILLISECONDS, WaitingConfig.timeout(), "Waiting for old buildConfigs deletion before creating new one").waitFor();
```
mchoma: same

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -107,7 +110,12 @@ public void update(OpenShift openShift) {
...
+		new SimpleWaiter(() -> openShift.getPods().stream().noneMatch(p -> podName.equals(p.getMetadata().getName())), TimeUnit.MILLISECONDS, WaitingConfig.timeout(), "Waiting for old pods deletion before creating new one").waitFor();
```
mchoma: same

mchoma:
Looks good to me.

PR #331, opened 2020-01-16
[closes #265] Discovery of Openshift binary client on OCP 4

### LINKED ISSUE #265: Logic for finding OC client does not work on OCP 4.1
1. Seems location for OCP 4.1 clients changed to 
`https://mirror.openshift.com/pub/openshift-v4/clients/oc`. Xtf use hardcoded url [1]

2. Seems OCP 4.1  url for finding out version changed from `/version/openshift` to `/version` . Xtf use hardcoded [2]

Xtf should somehow deal with that. Either have different logic for OCP 3.x and OCP 4.x or just externalize these parameters with `-Dxtf.*` properties.

```
curl -k https://api.eap-qe-mnovak2-ocp41.eap-qe-mnovak2-ocp41.fw.rhcloud.com:6443/version
{
  "major": "1",
  "minor": "13+",
  "gitVersion": "v1.13.4+81fc896",
  "gitCommit": "81fc896",
  "gitTreeState": "clean",
  "buildDate": "2019-04-21T23:18:54Z",
  "goVersion": "go1.11.5",
  "compiler": "gc",
  "platform": "linux/amd64"
}
```

[1] https://github.com/xtf-cz/xtf/blob/e7f024ff66245784eaca4f053544c6a11e360b05/core/src/main/java/cz/xtf/core/openshift/OpenShifts.java#L27
[2] https://github.com/xtf-cz/xtf/blob/e7f024ff66245784eaca4f053544c6a11e360b05/core/src/main/java/cz/xtf/core/openshift/OpenShifts.java#L95

llowinge:


llowinge:
Now it should work as follows:

- User provides Openshift version via parameter -> `oc` will be downloaded from https://mirror.openshift.com/pub/openshift-v4/clients/oc
- User doesn't provide Openshift version -> the lookup downloads `oc` directly from the Openshift (available from Openshift 4.3)

core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
```diff
@@ -183,22 +203,49 @@ private static String downloadOpenShiftBinary(String version) {
 		File ocTarFile = new File(workdir, "oc.tar.gz");
 		File ocFile = new File(workdir, "oc");
 
+		final String ocUrl = clientLocation + ocFileName;
 		try {
-			URL requestUrl = new URL(clientLocation + "oc.tar.gz");
-			FileUtils.copyURLToFile(requestUrl, ocTarFile, 20_000, 300_000);
+			URL requestUrl = new URL(ocUrl);
+
+			Https.copyHttpsURLToFile(requestUrl, ocTarFile, 20_000, 300_000);
```
maschmid: I don't think we should be using the trust-all https downloader (which is used for tests) to actually download binaries from internet.

core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
```diff
@@ -169,11 +170,35 @@ private static String getHomeDir() {
...
+				clientLocation = String.format(OCP4_CLIENTS_URL + "%s/%s/", version, systemType);
+			} else {
+				if (SystemUtils.IS_OS_MAC) {
+					systemType = "mac";
+					ocFileName = "oc.zip";
+				} else {
+					systemType = "linux";
+					ocFileName = "oc.tar";
+				}
+				final Optional<ModelNode> downloadsNodeOptional = apiResources("route.openshift.io/v1/routes").get("items").asList().stream().filter(m -> "downloads".equals(m.get("metadata").get("name").asString())).findAny();
```
mchoma: Can `Openshift.getRoutes()` be used here?

core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
```diff
@@ -183,22 +208,53 @@ private static String downloadOpenShiftBinary(String version) {
...
+		final String ocp3UrlVersion = OpenShiftConfig.url() + "/version/openshift";
+		if (Https.getCode(ocp3UrlVersion) == 200) { // for OCP 3
+			String content = Https.httpsGetContent(ocp3UrlVersion);
+			return ModelNode.fromJSONString(content).get("gitVersion").asString().replaceAll("^v(.*)", "$1");
+		} else { // for OCP version > 3
+			return apiResources("config.openshift.io/v1/clusterversions/version").get("status").get("desired").get("version").asString();
+		}
+	}
+
+	private static ModelNode apiResources(final String apiRequest) {
```
mchoma: Can method `apiResources(String apiRequest)` be moved to `OpenShift` as a general access method for api which is not covered otherwise. What about calling it in general way E.g. OpenShift.getResources(string resourceIdentifier)

Also can this be implemented in this way `Http.get(OpenShiftConfig.url() + "/apis/network.openshift.io/v1/hostsubnets").trustAll().bearerAuth(getMasterToken()).execute().response();` it seams shorter and readable

core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
llowinge: @mchoma Yes, thx for the tip

core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
```diff
@@ -183,22 +209,47 @@ private static String downloadOpenShiftBinary(String version) {
...
 
 	public static String getVersion() {
-		String content = Https.httpsGetContent(OpenShiftConfig.url() + "/version/openshift");
-		return ModelNode.fromJSONString(content).get("gitVersion").asString().replaceAll("^v(.*)", "$1");
+		final String ocp3UrlVersion = OpenShiftConfig.url() + "/version/openshift";
+		if (Https.getCode(ocp3UrlVersion) == 200) { // for OCP 3
+			String content = Https.httpsGetContent(ocp3UrlVersion);
+			return ModelNode.fromJSONString(content).get("gitVersion").asString().replaceAll("^v(.*)", "$1");
+		} else { // for OCP version > 3
+			final CustomResourceDefinitionContext crdContext = new CustomResourceDefinitionContext.Builder()
```
mchoma: Interesting, havent realize this can be written in this way. It is not custom resource. But as far as it works I am fine with this. It is more readable than it used to be.

PR #345, opened 2020-02-14
consider xtf.cz/keep=true label as valid for protecting resources fro…

simkam:
…m delete

I'm not confident about this change, but I guess I'm not the only one who tried `addToLabels(KEEP_LABEL, "true")` and spent time to investigate why it doesn't work. It is used even in xtf [OpenShift.java#L194](https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShift.java#L194) (it works there thanks to additional type check).

- [X] Pull Request contains a description of the changes
- [X] Pull Request does not include fixes for multiple issues/topics
- [X] Code is formatted, imports ordered, code compiles and tests are passing
- [X] Code is self-descriptive and/or documented

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -533,7 +533,7 @@ public class OpenShift extends DefaultOpenShiftClient {
 	 * @return List of secrets that aren't considered default.
 	 */
 	public List<Secret> getUserSecrets() {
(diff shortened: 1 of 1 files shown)


mnovak1:
LGTM. 

@simkam do you remember if there is any run with EAP Openshift TS with this change?

simkam:
@mnovak1 no, I don't remember

mnovak1:
@simkam Ok, change appears to be ok so merging. But it will be re-tested with next XTF upgrade into EAP Openshift test suite.

PR #357, opened 2020-05-15
WIP: Adding support for tracing imagestreams creation

fabiobrz:
This PR is to add a new class for image streams operations in order to keep track of those created through a `ConcurrentHashMap` that would serve as a cache.
This is to avoid errors when requesting for the creation of an image stream that already exists in the namespace.

The PR is WIP because there seem to be a specific case that should still be managed: when the image stream creation is requested by some internal operation (e.g.: some operator code) the newly introduced cache wouldn't be aware of it, becoming unreliable.

@istraka I'd like to assign this PR review to you but it seems I am not able to do that.
CC @mnovak1 
 
### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -2,6 +2,8 @@ package cz.xtf.core.openshift;
 
 import cz.xtf.core.config.WaitingConfig;
 import cz.xtf.core.openshift.crd.CustomResourceDefinitionContextProvider;
+import cz.xtf.core.openshift.client.dsl.TracedImageStreamOperations;
+import cz.xtf.core.openshift.imagestreams.MapBasedImageStreamRegistry;
 import cz.xtf.core.waiting.SimpleWaiter;
 import cz.xtf.core.waiting.Waiter;
 import io.fabric8.kubernetes.api.model.ConfigMap;
@@ -32,15 +34,19 @@ import io.fabric8.kubernetes.api.model.rbac.RoleBindingBuilder;
 import io.fabric8.kubernetes.api.model.rbac.Subject;
 import io.fabric8.kubernetes.api.model.rbac.SubjectBuilder;
 import io.fabric8.kubernetes.client.Config;
-import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.LocalPortForward;
+import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.dsl.LogWatch;
+import io.fabric8.kubernetes.client.dsl.MixedOperation;
+import io.fabric8.kubernetes.client.dsl.Resource;
 import io.fabric8.openshift.api.model.Build;
 import io.fabric8.openshift.api.model.BuildConfig;
 import io.fabric8.openshift.api.model.BuildRequest;
 import io.fabric8.openshift.api.model.BuildRequestBuilder;
 import io.fabric8.openshift.api.model.DeploymentConfig;
+import io.fabric8.openshift.api.model.DoneableImageStream;
 import io.fabric8.openshift.api.model.ImageStream;
+import io.fabric8.openshift.api.model.ImageStreamList;
 import io.fabric8.openshift.api.model.Project;
 import io.fabric8.openshift.api.model.ProjectRequest;
 import io.fabric8.openshift.api.model.ProjectRequestBuilder;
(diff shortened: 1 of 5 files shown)


istraka:
FYI @simkam @mchoma

istraka:
@fabiobrz  Does `handle create operation` fail if image stream is created by operator? 
If no, do we have to care? Same image stream would be re-imported, added to cache and metadata would be available.

fabiobrz:
> @fabiobrz Does `handle create operation` fail if image stream is created by operator?

Hi @istraka. I have not tried this specific case (error) yet but I've seen that an image stream created by an operator would not be traced by the cache, because it is in server scope.
 This implementation, but I'd say the whole project, is about the client. If an operator or an internal piece of code creates a resource, then the client workflow is not aware of that, ATM. 
There could be some way to do this but IMO those would suffer from with latency - i.e. polling with `get()`s as there's no concurrency regulation between client and server scope.

> If no, do we have to care? Same image stream would be re-imported, added to the registry and metadata would be available.

That's the point, IMHO. The XTF project, IIUC, is about client scope - i.e. the same scenario that you've just mentioned could take place even without these changes. 

Nevertheless, I've some changes to be pushed to take care of another use case, too - i.e. deletion of a shared image stream (one creates it, another one relies on it, then the first one shouldn't be able to actually delete it if the resource is still used by the second requester). For this I introduced subscriptions.
I'll push the change in a few minutes.

fabiobrz:
Hi @istraka, changes are pushed but I was not able to test this because of cluster being not available (at least to me...).
Anyway this is still WIP as the `handleReplace()` case should be managed yet.
Let me know what you think, I am glad to take part in this discussion.

mchoma:
`This is to avoid errors when requesting for the creation of an image stream that already exists in the namespace.`
I would say mechanism you are suggesting is new in XTF. Which smells to me. Question I ask is how we did until know? Because this is general problem of all Kubernetes resources, not just image streams. 

I think we do not hit this problem often, because tests works on clean namespace mostly.
Or we can leverage `createOrReplace` method instead of `create`, e.g. `secrets().createOrReplace(pullSecret);`
Or there is just check `if (doesNotExist) create`.

So why is your case specific?

fabiobrz:
Hi @mchoma,

> I would say mechanism you are suggesting is new in XTF. Which smells to me. Question I ask is how we did until know? Because this is general problem of all Kubernetes resources, not just image streams.

I'll start from the last one: you're right on that. I've been discussing about a "generic-oriented" solution in order to extend the behavior to all of the k8 resources but it was actually out of scope with respect to the task I was assigned.
That said, generally speaking and in order to answer your first question, I'd say that I don't know how you've done it until now. 
The matter here is that the solution I am proposing was about avoiding errors when asking for creation of an image stream that already exists in the namespace. This occurred specifically when a race condition was triggered by requesting image metadata (and image stream creation consequently) from an annotation condition logic.
 
> I think we do not hit this problem often, because tests works on clean namespace mostly.
Or we can leverage `createOrReplace` method instead of `create`, e.g. `secrets().createOrReplace(pullSecret);`
Or there is just check `if (doesNotExist) create`.

yes, strictly speaking I agree with you. This one was supposed to be an embedded (meaning in XTF core scope) mechanism to grant for image stream creation concurrency issues to be avoided but yeah, from my PoV I must admit that this should be nothing else that client code business. That said, anyhow, I think that ATM the only way that clients can follow in order to grant a single access point to `operations` which extend default implementation would be to extend the `OpenShift` class, wouldn't it?

> So why is your case specific?

See above - i.e. annotation condition logic (https://github.com/xtf-cz/xtf/pull/358).

If we're all on the same page on the above reasoning I can close this and see whether a more loosely coupled solution can be achieved for the same goal. @istraka WDYT?

istraka:
```
Because this is general problem of all Kubernetes resources, not just image streams.
```
If we took this path we would need to make thread safe everything else (eventually). I see two ways: 
* do it in XTF - apply the same principle, which would bring enormous maintenance cost
* do it in Fabric8

Considering Martin proposal:
* `OpenShift` class encapsulate operations for many resources:
** `deleteService(Service service)`
** `createRoute(Route route)`
* Either

istraka:
```
Because this is general problem of all Kubernetes resources, not just image streams.
```
If we took this path we would need to make thread safe everything else (eventually). I see two ways: 
* do it in XTF - apply the same principle, which would bring enormous maintenance cost
* do it in Fabric8

Considering Martin proposal:
- `OpenShift` class encapsulate operations for many resources:
    - `deleteService(Service service)`
    - `createRoute(Route route)`
- Some of these methods are useful, some of them are just stupid encapsulation to maintain unified approach

With this approach we can simply add `createOrReplaceImageStream` method and analyze & refactor/change testsuite.

After considering issues with operators creating images, which is totally out of control, I am in favor of this loosely coupled solution. Moreover this approach may easily applied to another resources in contrast to overriding `MixedOperation`.

istraka:
This can be closed.

PR #358, opened 2020-05-15
Allow use of Docker image metadata in @SkipFor and @SinceVersion annotations in 'name'

fabiobrz:
This PR introduces a new class, namely `DockerImageMetadata` in `Junit5` module.
The class is performs to retrieval and caching of Docker image metadata and exposes them with convenience methods that are then used by the `@SkipFor` annotation - see `SkipForCondition` class - in order to selectively disable tests based on the image name. 
This is an alternative solution to the existing implementation which relies on regex.

The changes introduced by this PR are actually depending on https://github.com/xtf-cz/xtf/pull/357 because the image metadata retrievakl ends up in calling `OpenShift.createImageStream()` which would fail in case the requested image is already present in the given namespace.

@istraka I'd like to assign this PR review to you.
CC @simkam, @mchoma, @mnovak1.

### DIFF
--- a/junit5/src/main/java/cz/xtf/junit5/extensions/SkipForCondition.java
+++ b/junit5/src/main/java/cz/xtf/junit5/extensions/SkipForCondition.java
@@ -1,17 +1,16 @@
 package cz.xtf.junit5.extensions;
 
 import cz.xtf.core.image.Image;
+import cz.xtf.core.openshift.OpenShifts;
 import cz.xtf.junit5.annotations.SkipFor;
 import cz.xtf.junit5.annotations.SkipFors;
+import cz.xtf.junit5.model.DockerImageMetadata;
 import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.platform.commons.support.AnnotationSupport;
 
-import java.util.regex.Pattern;
-
 public class SkipForCondition implements ExecutionCondition {
-
 	@Override
 	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
 		SkipFor skipFor = AnnotationSupport.findAnnotation(context.getElement(), SkipFor.class).orElse(null);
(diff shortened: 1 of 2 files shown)


junit5/src/main/java/cz/xtf/junit5/extensions/SkipForCondition.java
```diff
@@ -32,8 +31,10 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
 
 	public static ConditionEvaluationResult resolve(SkipFor skipFor) {
 		Image image = Image.resolve(skipFor.image());
-		Pattern name = Pattern.compile(skipFor.name());
-		if (name.matcher(image.getRepo()).matches()) {
```
simkam, line 36: doesn't this break current uses of `name` with regexp ?

fabiobrz: Yep, although in the case I've been tackling there wasn't any such case and the regex mlogic was introduced just few days ago. 
But yeah, in general you're right. This should be conditional.

mchoma: We want to be able to use regexp in SkipFor against image name from metadata. It is not either or, we want both features in conjuction.

Now we have this 
```
	@SkipFor(image = "eap", name = "eap-xp1-openjdk8-openshift-rhel7", reason = "/opt/eap/standalone/deployments/activemq-rar.rar is not in xp image")
	@SkipFor(image = "eap", name = "jboss-eap-7-eap-xp1-openjdk8-openshift-rhel7", reason = "/opt/eap/standalone/deployments/activemq-rar.rar is not in xp image")
	@SkipFor(image = "eap", name = "eap-xp1-openjdk11-openshift-rhel8", reason = "/opt/eap/standalone/deployments/activemq-rar.rar is not in xp image")
	@SkipFor(image = "eap", name = "jboss-eap-7-eap-xp1-openjdk11-openshift-rhel8", reason = "/opt/eap/standalone/deployments/activemq-rar.rar is not in xp image")
	@Test
	public void testActiveMqRarVersion() throws IOException {
``` 

With comming xp2 it will double. We want to be able to have something like this for all xp versions

```
	@SkipFor(image = "eap", name = "*eap-xp*", reason = "/opt/eap/standalone/deployments/activemq-rar.rar is not in xp image")
	@Test
	public void testActiveMqRarVersion() throws IOException {
```

junit5/src/main/java/cz/xtf/junit5/extensions/SkipForCondition.java
```diff
@@ -32,8 +31,10 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
...
+		DockerImageMetadata metadata = DockerImageMetadata.prepare(OpenShifts.master(), image);
```
mchoma, line 34: What does `DockerImageMetadata.prepare` do underneath? I hope it is not creating image stream.

istraka:
This can be closed.

PR #359, opened 2020-05-21
add OpenshiftStateRecorder

mocenas:
Add OpenshiftStateRecorder class. Serves to store info about openshift state, in case of test or build/deployment failure.

Also updates junit-jupiter to version 5.6.2, because LifecycleMethodExecutionExceptionHandler requires at least 5.5.

[example-log.zip](https://github.com/xtf-cz/xtf/files/4676210/example-log.zip)


junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
```diff
@@ -0,0 +1,199 @@
...
+ *    Is called whenever test ends. Stores openshift state if test failed.
+ *    To use this interface add this class as a listener
+ *
+ * LifecycleMethodExecutionExceptionHandler
+ *    This interface is executed when @BeforeAll or @BeforeEach method fails - throw an exception.
+ *    To use this interface a test class needs to be extended with this class using @ExtendWith
+ */
+@Slf4j
+public class OpenshiftStateRecorder implements TestExecutionListener, LifecycleMethodExecutionExceptionHandler {
+		private static final Path statusDir = Paths.get("log","status");
```
Crumby, line 48: Why two tabs for space? All the project uses one tab everywhere...

mocenas: @Crumby Fixed.

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * LifecycleMethodExecutionExceptionHandler
+ *    This interface is executed when @BeforeAll or @BeforeEach method fails - throw an exception.
+ *    To use this interface a test class needs to be extended with this class using @ExtendWith
+ */
+@Slf4j
+public class OpenshiftStateRecorder implements TestExecutionListener, LifecycleMethodExecutionExceptionHandler {
+	private static final Path statusDir = Paths.get("log","status");
+
+	private final OpenShift openShift = OpenShifts.master();
```
mchoma, line 50: could you move initialisation into constructor to be consistent with rest?

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
```diff
@@ -0,0 +1,199 @@
...
+		for (DeploymentConfig deploymentConfig : openShift.getDeploymentConfigs()) {
+			writer.write(deploymentConfig.getMetadata().getName()+ " replicas:" + deploymentConfig.getStatus().getReplicas() + " readyReplicas: " + deploymentConfig.getStatus().getReadyReplicas());
+			writer.newLine();
+		}
+		writer.newLine();
+
+
+		writer.write("Events:");
+		writer.newLine();
+		for (Event event : openShift.getEvents()){
```
mchoma, line 164: This will get all events in past N hours. N is cluster wide configurable, typically 2 hours.

Is it possible to filter just events which are connected to current test? E.g. by timestamp filtering?

If events will be not filtered just to current test, then it does not add much. There are already events saved once testsuite finished - by `EventsRecorder`.

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
```diff
@@ -0,0 +1,199 @@
...
+		if (!namespace.equals(buildNamespace)){
+			storeNamespaceState(testLogDir, buildOpenShift);
+			storePods(testLogDir, buildOpenShift);
+		}
+	}
+
+	private void storePods(Path dir, OpenShift openshift){
+		for (Pod pod : openshift.getPods()) {
+			try {
+				openshift.storePodLog(pod, dir, pod.getMetadata().getName() + ".log");
```
mchoma, line 121: In xtf there is already `PodLogRecorder` to store logs once test failed. 

What does it adds on top of that?

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	private void storeNamespaceState(Path directory, OpenShift openShift) throws IOException {
+		String namespace = openShift.getNamespace();
+		BufferedWriter writer = new BufferedWriter(new FileWriter(directory.toAbsolutePath() + "/namespace-" + namespace + ".txt"));
+		writer.write("Namespace " + namespace + " state - " + LocalDateTime.now());
+		writer.newLine();
+
+		writer.write("Pods: (name ; status ; readyContainers/totalContainers ; containerRestart)");
```
mchoma, line 136: I think this is useful information. We need probably more stuff like that.

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
```diff
@@ -0,0 +1,199 @@
...
+			ex.addSuppressed(newEx);
+		}
+		throw ex;
+	}
+
+	// Log the openshift state if test fail
+	@Override
+	public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult testExecutionResult) {
+		try {
+			if (FAILED.equals(testExecutionResult.getStatus()) && testIdentifier.isTest()) {
```
mchoma, line 97: Could that be configurable, so I am able to store the state also for PASSED test, e.g. with something like `xtf.config.store_state_always` ?

By default it will store infor just for failed tests.

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
```diff
@@ -0,0 +1,199 @@
...
+				storeOpenshiftState(getTestDisplayName(testIdentifier));
+			}
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
+	}
+
+	private void storeOpenshiftState(String testIdentifier) throws IOException {
```
mchoma, line 105: I can imagine it could be useful to have method like this available in XTF public API. So when troubelshooting some test I can let store the state in some important points. E.g. before test and after test or some important place in the middle.

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
mocenas: Nice catch I will look into this. Original purpose for this was to easily find out, when tests fails because of some image pull failure etc. 
Aren't events are also limited by number? I think I never got more that about 100, but I'm not sure about this.

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
mocenas: Nothing from pod logs themselfs. Why I added it to this is:
- It also records from build namespace.
- It is also recorded when Before* method fails. So it's possible to get logs from builds, non-tested services and other thing that can fail in setup.

junit5/src/main/java/cz/xtf/junit5/listeners/OpenshiftStateRecorder.java
mocenas: Yes it could be worth it. 
Any idea what is the right place to put it in? I think that Openshift class might be good, but not sure if it fits here.

mocenas:
> I think overal this is good idea. We need to identify useful information to store. Could you also attach example of information stored?

Yes example log attached.

mchoma:
@mnovak1 @istraka I know you are thinking about something similar. Can you drop comment here how it relates to this MR?

istraka:
@mchoma yes, you are correct. We need to record isolated logs for test. And we need to consider parallelism hence we can record everything in a namespace. I went through this PR and toke it as inspiration for some features: record always, record services, secrets,...However there will be some other differences.

mnovak1:
@istraka @mocenas Can be this one closed? I think this PR was incorporated in https://github.com/xtf-cz/xtf/pull/365, right?

istraka:
@mnovak1 yes for me.

mocenas:
@mnovak1 Yes it can be closed.

PR #362, opened 2020-06-01
Provide robust solution for recognizing image name for SkipFor logic

istraka:
**DockerImageMetadata**
- does not depend on jboss.dmr
- caches metadata for image streams
- checks for existing ones
- thread safe
- same api as ImageMetadata (means to replace it)

**SkipForCondition**
- uses DockerImageMetadata class
- gets name from metadata and applies the pattern

ImageMetadata class is deprecated now

### DIFF
--- a/junit5/src/main/java/cz/xtf/junit5/annotations/SinceVersion.java
+++ b/junit5/src/main/java/cz/xtf/junit5/annotations/SinceVersion.java
@@ -15,10 +15,21 @@ import java.lang.annotation.Target;
 @ExtendWith(SinceVersionCondition.class)
 public @interface SinceVersion {
 
+
 	/**
 	 * Name or regexp pattern matching name of the image. For example "eap73-openjdk11-openshift-rhel8" or "eap73-openjdk11-.*"
+	 * <p>
+	 * Only one of {@code imageMetadataLabelName} and {@code name} can be presented.
 	 */
-	String name();
+	String name() default "";
+
+	/**
+	 * Name or regexp pattern matching name in {@code Docker Labels} in image metadata
+	 * For example "eap73-openjdk11-openshift-rhel8" or "eap73-openjdk11-.*".
+	 * <p>
+	 * Only one of {@code imageMetadataLabelName} and {@code name} can be presented.
+	 */
+	String imageMetadataLabelName() default "";
 
 	String image();
 
--- a/junit5/src/main/java/cz/xtf/junit5/annotations/SkipFor.java
+++ b/junit5/src/main/java/cz/xtf/junit5/annotations/SkipFor.java
@@ -17,8 +17,18 @@ public @interface SkipFor {
 
 	/**
 	 * Name or regexp pattern matching name of the image. For example "eap73-openjdk11-openshift-rhel8" or "eap73-openjdk11-.*"
+	 * <p>
+	 * Only one of {@code imageMetadataLabelName} and {@code name} can be presented.
 	 */
-	String name();
+	String name() default "";
+
+	/**
+	 * Name or regexp pattern matching name in {@code Docker Labels} in image metadata
+	 * For example "eap73-openjdk11-openshift-rhel8" or "eap73-openjdk11-.*".
+	 * <p>
+	 * Only one of {@code imageMetadataLabelName} and {@code name} can be presented.
+	 */
+	String imageMetadataLabelName() default "";
 
 	String image();
 
--- a/junit5/src/main/java/cz/xtf/junit5/extensions/SinceVersionCondition.java
+++ b/junit5/src/main/java/cz/xtf/junit5/extensions/SinceVersionCondition.java
@@ -1,13 +1,16 @@
 package cz.xtf.junit5.extensions;
 
 import cz.xtf.core.image.Image;
+import cz.xtf.core.openshift.OpenShifts;
 import cz.xtf.junit5.annotations.SinceVersion;
 import cz.xtf.junit5.annotations.SinceVersions;
+import cz.xtf.junit5.model.DockerImageMetadata;
 import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.platform.commons.support.AnnotationSupport;
 
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class SinceVersionCondition implements ExecutionCondition {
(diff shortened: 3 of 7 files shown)


istraka:
@rjanik @mchoma @mnovak1 @simkam

junit5/src/main/java/cz/xtf/junit5/extensions/SkipForCondition.java
```diff
@@ -32,8 +33,11 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
 
 	public static ConditionEvaluationResult resolve(SkipFor skipFor) {
 		Image image = Image.resolve(skipFor.image());
+		DockerImageMetadata metadata = DockerImageMetadata.get(OpenShifts.master(), image);
+		String[] imageNameParts = metadata.labels().get("name").split("/");
+		String actualImageName = imageNameParts.length == 2 ? imageNameParts[1] : imageNameParts[0];
 		Pattern name = Pattern.compile(skipFor.name());
-		if (name.matcher(image.getRepo()).matches()) {
+		if (name.matcher(actualImageName).matches()) {
```
LittleJohnII, line 40: I believe this is backwards incompatible, right? We're still allowing regexes, but now you're checking against the `name` label instead of checking against the image name. All consumers of `@SkipFor` would have to change the `name` in the annotations now.

I propose to either:
* Deprecate SkipFor and create a new annotation which will check against metadata (or will allow both metadata and image name checks).
* Make the `name` parameter in `@SkipFor` optional, and add another parameter for metadata parsing. Only one can be defined at a time and the implementation will check against what's defined.

istraka: Yes, you are right. I went with the second option. You can check it out. I will let you know once I will finish test.

junit5/src/main/java/cz/xtf/junit5/annotations/SkipFor.java
```diff
@@ -16,9 +16,20 @@
 public @interface SkipFor {
 
 	/**
-	 * Name or regexp pattern matching name of the image. For example "eap73-openjdk11-openshift-rhel8" or "eap73-openjdk11-.*"
+	 * Name or regexp pattern matching name of the image.
+	 * For example "eap73-openjdk11-openshift-rhel8" or "eap73-openjdk11-.*".
+	 *
+	 * One of {@code dockerLabelName} or {@code name} must be present.
```
mnovak1, line 22: I think here should be provided name of the property from which the name is taken so it's clear what to specify here.

junit5/src/main/java/cz/xtf/junit5/annotations/SkipFor.java
```diff
@@ -16,9 +16,20 @@
...
-	String name();
+	String name() default "";
+
+	/**
+	 * Name or regexp pattern matching name of the docker image (docker labels -> name).
+	 * For example "eap73-openjdk11-openshift-rhel8" or "eap73-openjdk11-.*".
+	 *
+	 * One of {@code dockerLabelName} or {@code name} must be present.
+	 */
+	String dockerLabelName() default "";
```
mnovak1, line 32: The same here. Could you paste here example name from labels, please?

junit5/src/main/java/cz/xtf/junit5/extensions/SkipForCondition.java
mnovak1: The backward compatibility might be guaranteed also by matching `name` from `@SkipFor` against both:
-  image.get()
- actualImageName

If any of the them matches then test is skipped and avoid adding new property into `@SkipFor` annotation, wdyt?

istraka: Yeah. It changes old functionality, however keep backward compatibility to some level. Strictly speaking, it may results to skipping tests that should not be skipped. However I don't see this is a real scenario. Image name is very similar to docker name label.

junit5/src/main/java/cz/xtf/junit5/annotations/SkipFor.java
istraka: Please check.

junit5/src/main/java/cz/xtf/junit5/annotations/SkipFor.java
istraka: Please check.

junit5/src/main/java/cz/xtf/junit5/extensions/SkipForCondition.java
LittleJohnII: I'm not sure if I like it. I think it's a shady way of changing the behavior. When new people start writing tests, they'll see that in most existing tests the regexp matches `image.getRepo()`. They'll likely go on thinking that's the expected behavior and won't know that it also checks the label. Yes, it's documented, but people don't read documentation that often.

Additionally, there may be differences between the two, and adding this behavior may break existing tests. It's unlikely but it can happen. I'd still recommend one of my suggested solutions. Let's see what the others think.

junit5/src/main/java/cz/xtf/junit5/annotations/SinceVersion.java
```diff
@@ -15,10 +15,21 @@
...
+	String name() default "";
+
+	/**
+	 * Name or regexp pattern matching name in {@code Docker Labels} in image metadata
+	 * For example "eap73-openjdk11-openshift-rhel8" or "eap73-openjdk11-.*".
+	 * <p>
+	 * Only one of {@code dockerImageName} and {@code name} can be presented.
 	 */
-	String name();
+	String dockerImageName() default "";
```
mnovak1, line 32: `dockerImageName` does directly provide info what it is. What about to rename it to something like: `ImageMetadataLabelName`, wdyt?

junit5/src/main/java/cz/xtf/junit5/extensions/SinceVersionCondition.java
```diff
@@ -31,9 +34,22 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
...
 		Image image = Image.resolve(sinceVersion.image());
-		Pattern name = Pattern.compile(sinceVersion.name());
-		if (name.matcher(image.getRepo()).matches()) {
+		Matcher matcher;
+
+		if (!sinceVersion.name().equals("")) {
+			matcher = Pattern.compile(sinceVersion.name()).matcher(image.getRepo());
+		} else {
+			DockerImageMetadata metadata = DockerImageMetadata.get(OpenShifts.master(), image);
+			String[] imageNameParts = metadata.labels().get("name").split("/");
```
mnovak1, line 47: I think we should not take part after `/` in label name. In case that `name` is:
```"name": "jboss-eap-7/eap73-openjdk8-openshift-rhel7"``` 

then let's take the whole string. Otherwise I think it's confusing. Taking part after `/` had sense only in case where `name` and `dockerImageName` were the same.

LittleJohnII: Agreed, the part before the `/` might be important (if we want to filter out tech preview images for example).

LittleJohnII:
Looks good to me now, though Mirek's comments should still be addressed.

junit5/src/main/java/cz/xtf/junit5/extensions/SinceVersionCondition.java
istraka: The first part is repository. The second one is actual image name. I see your point, it might be usefull. With @mnovak1 proposal about changing the parameter name (`ImageMetadataLabelName`) it should be pretty straightforward.

istraka:
Fixed

mnovak1:
Merged, there was somehow stuck CI thus I did checks manually and passed.

Issue #340, opened 2020-01-31
More intelligent waiters


mnovak1:
Add waiter ability to wait for more complex events than pod up and pod down + implementation for the more common use cases.

Add waiter ability to fail fast, once a certain erroneous state has been detected (idea: lambdas for detection of error states, so it's configurable).

Crumby:
> Add waiter ability to wait for more complex events than pod up and pod down + implementation for the more common use cases.

Which complex events and common use cases do you have in mind particularly? 

> Add waiter ability to fail fast, once a certain erroneous state has been detected (idea: lambdas for detection of error states, so it's configurable).

This feature is already implemented by both [SimpleWaiter](https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/waiting/SimpleWaiter.java) and [SupplierWaiter](https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/waiting/SupplierWaiter.java). You can set it dynamically in the former. Later has to be created with the failure condition passed in constructor by design atm.

Check out [isDcReady()](https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java#L121) and [hasBuildCompleted()](https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java#L36) waiters as they both are complex and both contain fail fast ability. I'm open to implementing other if their use case will be general enough.

LittleJohnII:
@istraka with the fail-fast waiter support merged, can this be marked as done?

fabiobrz:
> @istraka with the fail-fast waiter support merged, can this be marked as done?

+1, can we close this as done? The PR for it mentions the related issue should be fixed, but unfortunately that was in [a comment](https://github.com/xtf-cz/xtf/pull/366#issuecomment-646713620) by the author, rather than in the PR description.

CC @mnovak1

mnovak1:
Closing.

PR #366, opened 2020-06-19
Redesign FailFast support

istraka:
Redesign FailFast support
- Interface FailFastCheck instead of BooleanSupplier which enables us to effectively store a reason of failure
- SupplierFailFastCheck - general implementation of FailFastCheck where we can simply put lambdas buildt in builders
- MultipleFailFastChecksHandler - class that implement FailFastCheck and handles list of checks - checks every one of them and if one fails provide the reason of that one
- EventFailFastCheckBuilder refactored EventFailFastBoolSupplierBuilder class - now it builds SupplierFailFastCheck and provides reason
- FailFastBuilder - refactored FailFastBoolSupplierBuilder

Classes are in new failfast package


istraka:
Fixes https://github.com/xtf-cz/xtf/issues/340.

core/src/main/java/cz/xtf/core/waiting/failfast/SupplierFailFastCheck.java
```diff
@@ -0,0 +1,45 @@
...
+	private final Function<X, String> reasonSupplier;
+
+	private String reason;
+
+	/**
+	 * @param supplier       supplier that provide object
+	 * @param shouldFail     condition. The result is the result of {@link FailFastCheck#shouldFail()}
+	 * @param reasonSupplier returns string as a reason why {@code shouldFail} returned {@code true}
+	 */
+	SupplierFailFastCheck(Supplier<X> supplier, Function<X, Boolean> shouldFail, Function<X, String> reasonSupplier) {
```
mnovak1, line 23: `supplier` is somewhat generic name of param and javadoc does not help much what it is. What about `watchedResourcesSupplier`

core/src/main/java/cz/xtf/core/waiting/failfast/SupplierFailFastCheck.java
```diff
@@ -0,0 +1,45 @@
...
+		this.eventSupplier = supplier;
+		this.shouldFail = shouldFail;
+		this.reasonSupplier = reasonSupplier;
+	}
+
+	@Override
+	public boolean shouldFail() {
+		X resources = eventSupplier.get();
+		if (shouldFail.apply(resources)) {
+			reason = reasonSupplier.apply(resources);
```
mnovak1, line 33: I'll need to see this in action. How does `reasonSupplier` looks like?

istraka: Check this out https://github.com/xtf-cz/xtf/pull/366/files#diff-1798cb6edc311f5fdc997027b5a83e6cR86

istraka:
Issues have been fixed.

core/src/main/java/cz/xtf/core/waiting/failfast/SupplierFailFastCheck.java
istraka: Yeah, sounds better.

core/src/main/java/cz/xtf/core/waiting/failfast/SupplierFailFastCheck.java
mnovak1: Ok, now I get it :-) it's a little bit different approach than I thought. In case we have number of ...`ofMessages(...)` message events based on which it failed fast then it would be great know exactly which it was. I think that then there should be added `FailFastCheck` for each event message separately with given condition to know which it was exactly, right?

mnovak1: I'm not saying that it should be changed, this is probably the only way how to avoid of setting reason directly in condition. I see it as tradeoff and when creating fail fast it will need to be basically multiple failfast.

istraka: Yeah, with current functionality it would need to be in that way (mupltiple FailFastChecks). But do we need to know which exact regex (what we are looking for) matched? The way I see it:
- we know what events show bad state
- when a bad state occurs I see list of events (in a exception) that is subset of those known events

Check this out: https://gitlab.mw.lab.eng.bos.redhat.com/jbossqe-eap/openshift-eap-tests/merge_requests/615#note_105572

mnovak1: Ok, I think it's ok

mnovak1:
@mchoma could you take a look as well, please?

istraka:
The PR is ready.

mchoma:
> @mchoma could you take a look as well, please?

I am not following anymore. I see you checked it so I think it is enough.

LittleJohnII:
Squashing and merging, thanks.

PR #370, opened 2020-07-03
Update to F8 4.10.2

Crumby:
There are two issues with current Fabric8 version (4.9.2):
* It doesn't work with OCP 3.11 as it attempts to delete v2beta2 autoscalers during clean which leads to 404 fabric8 client errors.
* BinaryBuild delete step causes pods to end in terminating state indefinitely with OCP 4.5 and 4.6. This is fixed by setting up Background deletion policy. (Works out of the box with 4.10.2, needs to be setup with 4.9.2)

@mocenas 4.10.2 shouldn't have the issues with events as the API is kept.

Tested with OpenShift 3.11, 4.5 and 4.6


core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -108,10 +109,10 @@ public void update(OpenShift openShift) {
 
 	@Override
 	public void delete(OpenShift openShift) {
-		openShift.imageStreams().withName(is.getMetadata().getName()).cascading(false).withGracePeriod(0).delete();
-		openShift.buildConfigs().withName(bc.getMetadata().getName()).cascading(true).withGracePeriod(0).delete();
+		openShift.imageStreams().withName(is.getMetadata().getName()).withPropagationPolicy(DeletionPropagation.BACKGROUND).delete();
+		openShift.buildConfigs().withName(bc.getMetadata().getName()).withPropagationPolicy(DeletionPropagation.BACKGROUND).delete();
```
mnovak1, line 113: Is it semantically identical? I mean that image streams were deleted with `cascading(false)` and build config with `cascading(true)` however `....withPropagationPolicy(DeletionPropagation.BACKGROUND)` is the same for both.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -1269,7 +1269,7 @@ private Path storeLog(String log, Path dirPath, String fileName) throws IOExcept
 	// Waiting
 
 	/**
-	 * Use {@link OpenShiftWaiters#get(OpenShift, BooleanSupplier)} instead.
+	 * Use {@link OpenShiftWaiters#get(OpenShift, cz.xtf.core.waiting.failfast.FailFastCheck)} instead.
```
mnovak1, line 1272: thanks for updating it

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
Crumby: It's definitely not semantically identical. `Background` policy is more or less equal of `cascading` set to `true` while If we'd like to keep the code semantically identical we'd have to set deletion option of IS to `Orphan` [1]. I wanted to unify resource deletion approach in the method (if I recall correctly then the current implementation is a result of work-arounding previous versions of the client).

Anyway this code can be left out because `Background` policy is now automatically set from Fabric 4.10.0. Just a note that `cascading` method was deprecated.

[1] https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

mnovak1: Thanks for explanation. I cannot figure out where it might be an issue so marking as lgtm.

PR #371, opened 2020-07-10
Refactor BinaryBuild architecture and add option to build from file

istraka:
@simkam @pkremens 

### DIFF
--- a/core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
+++ b/core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
@@ -1,40 +1,12 @@
 package cz.xtf.core.bm;
 
-import static org.apache.commons.io.output.NullOutputStream.NULL_OUTPUT_STREAM;
-
 import io.fabric8.kubernetes.api.model.DeletionPropagation;
-import org.apache.commons.codec.binary.Hex;
-import org.apache.commons.compress.archivers.ArchiveEntry;
-import org.apache.commons.compress.archivers.ArchiveException;
-import org.apache.commons.compress.archivers.ArchiveStreamFactory;
-import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
-import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
-import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.IOUtils;
-import org.apache.commons.io.filefilter.TrueFileFilter;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.PipedInputStream;
-import java.io.PipedOutputStream;
-import java.nio.file.Files;
 import java.nio.file.Path;
-import java.security.DigestOutputStream;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.time.Instant;
-import java.util.Collection;
 import java.util.Collections;
-import java.util.Date;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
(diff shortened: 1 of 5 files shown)


core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -56,7 +28,7 @@
 
 @Slf4j
 public abstract class BinaryBuild implements ManagedBuild {
-	private static final String CONTENT_HASH_LABEL_KEY = "xtf.bm/content-hash";
+	static final String CONTENT_HASH_LABEL_KEY = "xtf.bm/content-hash";
```
simkam, line 31: what is the reason for this change?

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -66,10 +38,10 @@
 	private final String builderImage;
 	private final Map<String, String> envProperties;
 
-	private final ImageStream is;
-	private final BuildConfig bc;
+	final ImageStream is;
```
simkam, line 41: Shouldn't it be `protected` so any subclass can see these fields?

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -81,26 +53,6 @@ public BinaryBuild(String builderImage, Path path, Map<String, String> envProper
 		this.bc = this.createBcDefinition();
 	}
 
-	@Override
-	public void build(OpenShift openShift) {
```
simkam, line 85: this breaks API, no? It should be at least replaced with abstract method

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -239,48 +191,5 @@ private BuildConfig createBcDefinition() {
...
-					}
-				}
-				o.closeArchiveEntry();
-			}
-			o.finish();
-		} catch (ArchiveException | IOException e) {
-			throw new RuntimeException(e);
-		}
-	}
+	abstract String getContentHash();
```
simkam, line 194: `protected` ?

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
istraka: Hash code is computed in subclasses now. I will change it to `protected`.

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
istraka: Yeah good catch. If an object typed as `BinaryBuld` instead of `ManagedBuild` then yes.

istraka:
@simkam I have fixed issues. Moreover I made all abstract method protected so they are is easily overridden.

core/src/main/java/cz/xtf/core/bm/BinaryBuild.java
```diff
@@ -238,49 +192,4 @@ private BuildConfig createBcDefinition() {
...
-				contentHash = Hex.encodeHexString(dos.getMessageDigest().digest()).substring(0, 63);
-			} catch (NoSuchAlgorithmException e) {
-				throw new RuntimeException(e);
-			}
-		}
-
-		return contentHash;
-	}
-
-	private void writeProjectTar(OutputStream os) {
```
mnovak1, line 260: `BinaryBuild` is abstract class inherited in `BinaryDockerBuild`, `BinarySourceBuild` and `MutableBinarySourceBuild`. I don't like that if `writeProjectTar` and  `getContentHash` are removed here in `BinaryBuild` then it duplicated in inhereted classes. I'm not sure what is the reason for change.

istraka: It has been moved into `BinaryBuildFromSources`.

mnovak1: Ok, got it.

PR #375, opened 2020-07-20
FailFast for DockerImageMetadata class 

istraka:
when pulling image - check for tag condition that indicate import failure

### DIFF
--- a/junit5/src/main/java/cz/xtf/junit5/model/DockerImageMetadata.java
+++ b/junit5/src/main/java/cz/xtf/junit5/model/DockerImageMetadata.java
@@ -4,8 +4,11 @@ import cz.xtf.core.image.Image;
 import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.waiting.SimpleWaiter;
 import cz.xtf.core.waiting.Waiter;
+import cz.xtf.core.waiting.failfast.FailFastCheck;
 import io.fabric8.openshift.api.model.ImageStream;
 import io.fabric8.openshift.api.model.ImageStreamTag;
+import io.fabric8.openshift.api.model.NamedTagEventList;
+import io.fabric8.openshift.api.model.TagEventCondition;
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


test-helpers/src/main/java/cz/xtf/testhelpers/image/ImageMetadata.java
```diff
@@ -120,4 +124,42 @@ public String command() {
 
 		return result;
 	}
+
+	private static final class ImageStreamFailFastCheck implements FailFastCheck {
```
mnovak1, line 128: This inner class is used twice (same code), could it be separated into some "failfast" package and used from there.

istraka:
There was an issue in `ImageMetadata`. It is ready now.

@LittleJohnII `ImageMetadata` is deprecated so this is rather enhancement. The class can be even removed sometime.

junit5/src/main/java/cz/xtf/junit5/model/DockerImageMetadata.java
```diff
@@ -161,4 +165,42 @@ public String command() {
...
+		@Override
+		public boolean hasFailed() {
+			ImageStream imageStream = openShift.getImageStream(imageName);
+			if (imageStream == null) {
+				return false;
+			}
+			for (NamedTagEventList tag : imageStream.getStatus().getTags()) {
+				if (image.getTag().startsWith(tag.getTag())) {
+					for (TagEventCondition condition : tag.getConditions()) {
+						if (condition.getType().equals("ImportSuccess") && condition.getStatus().equals("False")) {
```
mnovak1, line 191: In case that image stream is "creation in progress" then `ImportSuccess` is undefined? I'm just worrying how it works and would like to avoid issues when imagestream creation is in progress and this value being `false`. It would fail instead of wait.

test-helpers/src/main/java/cz/xtf/testhelpers/image/ImageMetadata.java
istraka: It is very specific to the case. I would not even add it into `ImageMetadata` class, however someone might benefit from this. Even us, until `DockerImageMetadata` class is used. Moreover `ImageMetadata` is deprecated and after updating policy it can be removed.

junit5/src/main/java/cz/xtf/junit5/model/DockerImageMetadata.java
istraka: No. Such list (conditions) is empty when import is successful. Even during the operation.

mnovak1: ok, thanks

test-helpers/src/main/java/cz/xtf/testhelpers/image/ImageMetadata.java
istraka: So If you really insist, I can move it to somewhere else.

mnovak1: Will be handled as part of https://github.com/xtf-cz/xtf/issues/378

PR #376, opened 2020-07-22
Print DockerImageRepository in openshift recorder.

mocenas:
Change OpenshiftRecorder to print DockerImageRepository not PublicDockerImageRepository. PublicDockerImageRepository may not be present in imageStream, which causes the recorder to throw NPE and crash.

### DIFF
--- a/junit5/src/main/java/cz/xtf/junit5/extensions/helpers/ResourcesPrinterHelper.java
+++ b/junit5/src/main/java/cz/xtf/junit5/extensions/helpers/ResourcesPrinterHelper.java
@@ -88,7 +88,7 @@ public class ResourcesPrinterHelper<X> implements AutoCloseable {
 	private static LinkedHashMap<String, String> getISCols(ImageStream is) {
 		LinkedHashMap<String, String> map = new LinkedHashMap<>(2);
 		map.put("NAME", is.getMetadata().getName());
-		map.put("IMAGE REPOSITORY", is.getStatus().getPublicDockerImageRepository());
+		map.put("IMAGE REPOSITORY", is.getStatus().getDockerImageRepository());
 		return map;
 	}
 
--- a/junit5/src/main/java/cz/xtf/junit5/extensions/helpers/ResourcesPrinterHelper.java
+++ b/junit5/src/main/java/cz/xtf/junit5/extensions/helpers/ResourcesPrinterHelper.java
@@ -89,6 +89,7 @@ public class ResourcesPrinterHelper<X> implements AutoCloseable {
 		LinkedHashMap<String, String> map = new LinkedHashMap<>(2);
 		map.put("NAME", is.getMetadata().getName());
 		map.put("IMAGE REPOSITORY", is.getStatus().getDockerImageRepository());
+		map.put("PUBLIC IMAGE REPOSITORY", is.getStatus().getPublicDockerImageRepository());
 		return map;
 	}
 
@@ -197,7 +198,9 @@ public class ResourcesPrinterHelper<X> implements AutoCloseable {
 
 	private void row(String[] cols) {
 		for (int i = 0; i < cols.length; i++) {
-			maxLengths[i] = Math.max(maxLengths[i], cols[i].length());
+			if (cols[i] != null) {
+				maxLengths[i] = Math.max(maxLengths[i], cols[i].length());
+			}
 		}
 		rows.add(cols);
 	}


mocenas:
cc @istraka

junit5/src/main/java/cz/xtf/junit5/extensions/helpers/ResourcesPrinterHelper.java
```diff
@@ -88,7 +88,7 @@ private ResourcesPrinterHelper(Path file, Function<X, LinkedHashMap<String, Stri
 	private static LinkedHashMap<String, String> getISCols(ImageStream is) {
 		LinkedHashMap<String, String> map = new LinkedHashMap<>(2);
 		map.put("NAME", is.getMetadata().getName());
-		map.put("IMAGE REPOSITORY", is.getStatus().getPublicDockerImageRepository());
+		map.put("IMAGE REPOSITORY", is.getStatus().getDockerImageRepository());
```
mnovak1, line 91: what's the difference between `publicDockerImageRepository` and `dockerImageRepository`?

mocenas: @mnovak1 simply publicDockerImageRepository is a path that can be accessed from outside the cluster. Described in this [doc](https://docs.openshift.com/container-platform/4.5/rest_api/image_apis/imagestream-image-openshift-io-v1.html).

mnovak1: would it make sense to have it as separate column? I mean:
```
map.put("IMAGE REPOSITORY", is.getStatus().getDockerImageRepository());
map.put("PUBLIC_IMAGE REPOSITORY", is.getStatus().getPublicDockerImageRepository());
```

mocenas: @mnovak1 Possibly, but a problem is, the publicDockerImageRepository is not always present (more precisely I've never seen it), which makes the reporter to fail on NPE, that needs to be addressed too. So it is possible, but I don't think it is worth it.

mnovak1: @istraka ^ 

is there any reason why null cannot be handled? I would expect that in this case value is undefined.

istraka: Null is not expected. Workaround is to put empty string. I assume NPE comes from here https://github.com/xtf-cz/xtf/blob/master/junit5/src/main/java/cz/xtf/junit5/extensions/helpers/ResourcesPrinterHelper.java#L200 where `cols[i]` is `null`. Can be easily fixed. Can it be part of the mr @mocenas ?

 The function follows `oc get is` command, which print `publicDockerImageRepository`. Not sure what is the output when the field is `null`. What about printing `getDockerImageRepository` if `publicDockerImageRepository` is empty? Does it cover your usecase?

mocenas: @istraka Yes NPE comes from there, I will fix it.
Rather than choosing which docker image to print, I would print both. If `publicDockerImageRepository` is null it will just print null to the output. What do you think?

BTW: For me `oc get is` for me prints `dockerImageRepository`. But as I mentioned above, I've never seen `publicDockerImageRepository` in use.

mnovak1: > Yes NPE comes from there, I will fix it.

great.

> If publicDockerImageRepository is null it will just print null to the output.

+1

@mocenas I plan to do new release today or tomorrow, will you be able to make the changes soon?

mocenas: @mnovak1 The changes are already pushed.

mnovak1:
lgtm

junit5/src/main/java/cz/xtf/junit5/extensions/helpers/ResourcesPrinterHelper.java
istraka: @mocenas 
```
            "status": {
                "dockerImageRepository": "image-registry.openshift-image-registry.svc:5000/istraka-namespace-builds/ee-defaults-amq7-eap-cd",
                "publicDockerImageRepository": "default-route-openshift-image-registry.apps.eapqe-010-hlpj.dynamic.xpaas/istraka-namespace-builds/ee-defaults-amq7-eap-cd"
            }
```

this is `-o json` output for me. I've assumed that `publicDockerImageRepository` comes before `dockerImageRepository`.

PR #377, opened 2020-07-22
Store openshift state if before* fails.

mocenas:
Record openshift state even if beforeAll or beforeEach method fails.


istraka:
@mocenas can you add properties to control this functionality?

mocenas:
@istraka Added.

PR #403, opened 2021-01-11
extend ocConfigPath compatibility to OCP 4.7

fmongiar:

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
@@ -52,7 +52,7 @@ public class OpenShiftBinary {
         if (ocConfigPath == null) {
             return executeCommand(ArrayUtils.addAll(new String[] { path }, args));
         } else {
-            return executeCommand(ArrayUtils.addAll(new String[] { path, "--config=" + ocConfigPath }, args));
+            return executeCommand(ArrayUtils.addAll(new String[] { path, "--kubeconfig=" + ocConfigPath }, args));
         }
     }
 


fmongiar:
@mnovak1 could you please check this out? It is needed for lpinterop vs OCP4.7.
I tested this against:
- OCP 3.11.135
- OCP 4.6.9
- OCP 4.7.0-0.nightly-2020-12-14-165231

simkam:
To add some background. `--config` flag was deprecated and it is removed in 4.7.

* 4.5
```
$ ./oc version --client
Client Version: 4.5.24
5$ ./oc --config --help
Flag --config has been deprecated, use --kubeconfig instead
```
* 4.6
```
$ ./oc version --client
Client Version: 4.6.9
$ oc --config --help
Flag --config has been deprecated, use --kubeconfig instead
```
* 4.7
```
$ ./oc version --client
Client Version: 4.7.0-0.nightly-2021-01-10-070949
$ ./oc --config --help
Error: unknown flag: --config
```

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
```diff
@@ -52,7 +53,15 @@ public String execute(String... args) {
         if (ocConfigPath == null) {
             return executeCommand(ArrayUtils.addAll(new String[] { path }, args));
         } else {
-            return executeCommand(ArrayUtils.addAll(new String[] { path, "--config=" + ocConfigPath }, args));
+            log.debug("openshift version: [{}]", OpenShifts.getVersion());
+            if (StringUtils.isNotEmpty(OpenShifts.getVersion())) {
+                String[] ocpVersion = OpenShifts.getVersion().split("\\.");
+                if (Integer.parseInt(ocpVersion[0]) < 4
+                        || Integer.parseInt(ocpVersion[0]) == 4 && Integer.parseInt(ocpVersion[1]) < 6) {
```
simkam, line 60: To keep 4.6 behavior unchanged the last condition should be `< 7` or `<= 6`. Not a big problem as `--kubeconfig` works with 4.6.

simkam (approved):

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
fmongiar: should I change the condition then or it is fine like that?
I mean for OCP 4.6 the `--kubeconfig` is working as well as `--config`

simkam: It is fine for me, but let's wait for @mnovak1 (OOO today) or any other reviewer. I cannot merge it anyway.

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
```diff
@@ -52,7 +53,15 @@ public String execute(String... args) {
         if (ocConfigPath == null) {
             return executeCommand(ArrayUtils.addAll(new String[] { path }, args));
         } else {
-            return executeCommand(ArrayUtils.addAll(new String[] { path, "--config=" + ocConfigPath }, args));
+            log.debug("openshift version: [{}]", OpenShifts.getVersion());
```
mnovak1, line 56: Problem with `OpenShifts.getVersion()` is that it returns version of OCP cluster and not the `oc` client. There are two options, add logic to get oc client version or just use `--kubeconfig=` for all `oc` versions. I would prefer 2nd option as we don't need to mess up the code by IFs for given versions. wdyt?

simkam: by default, XTF downloads client version corresponding to cluster version, but yeah, it is issue when version is overriden in config.

fmongiar: I like the second option and it is compatible with 3.11 too. Going for update PR

mnovak1 (approved):

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
mnovak1: thanks

PR #407, opened 2021-01-15
[feature] Label every created namespace.

jstourac:
Motivation for this is to have some common label for the projects/namespaces created by XTF automatically. This can ease their listing and tracing.

So far we search based on the 'openshift.io/requester' annotation value.

### DIFF
--- a/core/src/main/java/cz/xtf/core/bm/BuildManager.java
+++ b/core/src/main/java/cz/xtf/core/bm/BuildManager.java
@@ -3,7 +3,9 @@ package cz.xtf.core.bm;
 import cz.xtf.core.config.BuildManagerConfig;
 import cz.xtf.core.config.OpenShiftConfig;
 import cz.xtf.core.openshift.OpenShift;
+import cz.xtf.core.openshift.OpenShifts;
 import cz.xtf.core.waiting.Waiter;
+import io.fabric8.kubernetes.client.KubernetesClientException;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
@@ -16,6 +18,24 @@ public class BuildManager {
         if (openShift.getProject(openShift.getNamespace()) == null) {
             openShift.createProjectRequest();
             openShift.waiters().isProjectReady().waitFor();
+
+            try {
+                // Adding a label can be only done via 'namespace'. It cannot be set via 'project' API. Thus we do this
+                // separately. Also, to update namespace label, it's necessary to have 'patch resource "namespaces"'
+                // permission for current user and updated namespace, e.g. by having 'cluster-admin' role.
+                // Otherwise you can see:
+                // $ oc label namespace <name> "label1=foo"
+                // Error from server (Forbidden): namespaces "<name>" is forbidden: User "<user>" cannot patch resource "namespaces" in API group "" in the namespace "<name>"
+                OpenShifts.admin().namespaces().withName(openShift.getNamespace()).edit().editMetadata().addToLabels(
+                        OpenShift.XTF_MANAGED_LABEL, "true").endMetadata().done();
+            } catch (KubernetesClientException e) {
+                // We weren't able to assign a label to the new project. Let's just print warning since this information
+                // is not critical to the tests execution. Possible cause for this are insufficient permissions since
+                // some projects using XTF are executed on OCP instances without 'admin' accounts available.
+                log.warn("Couldn't assign label '" + OpenShift.XTF_MANAGED_LABEL + "' to the new project '"
+                        + openShift.getNamespace() + "'. Possible cause are insufficient permissions.");
+                log.debug(e.getMessage());
+            }
         }
         if (OpenShiftConfig.pullSecret() != null) {
             openShift.setupPullSecret(OpenShiftConfig.pullSecret());
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -93,6 +93,12 @@ public class OpenShift extends DefaultOpenShiftClient {
     private static volatile String routeSuffix;
 
     public static final String KEEP_LABEL = "xtf.cz/keep";
+    /**
+     * This label is supposed to be used for any resource created by the XTF to easily distinguish which resources have
+     * been created by XTF automation.
+     * NOTE: at the moment only place where this is used is for labeling namespaces. Other usages may be added in the future.
+     */
+    public static final String XTF_MANAGED_LABEL = "xtf.cz/managed";
     private final AppsAPIGroupClient appsAPIGroupClient;
 
     /**
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -253,7 +253,7 @@ public class OpenShift extends DefaultOpenShiftClient {
     }
 
     /**
-     * Calls rectreateProject(namespace).
+     * Calls recreateProject(namespace).
      *
      * @see OpenShift#recreateProject(String)
      */
@@ -265,7 +265,7 @@ public class OpenShift extends DefaultOpenShiftClient {
      * Creates or recreates project specified by name.
      *
      * @param name name of a project to be created
-     * @return ProjectRequest instatnce
+     * @return ProjectRequest instance
      */
     public ProjectRequest recreateProject(String name) {
         return recreateProject(new ProjectRequestBuilder().withNewMetadata().withName(name).endMetadata().build());
@@ -274,7 +274,7 @@ public class OpenShift extends DefaultOpenShiftClient {
     /**
      * Creates or recreates project specified by projectRequest instance.
      *
-     * @return ProjectRequest instatnce
+     * @return ProjectRequest instance
      */
     public ProjectRequest recreateProject(ProjectRequest projectRequest) {
         boolean deleted = deleteProject(projectRequest.getMetadata().getName());
@@ -287,7 +287,7 @@ public class OpenShift extends DefaultOpenShiftClient {
     }
 
     /**
-     * Tries to retreive project with name 'name'. Swallows KubernetesClientException
+     * Tries to retrieve project with name 'name'. Swallows KubernetesClientException
      * if project doesn't exist or isn't accessible for user.
      *
      * @param name name of requested project.
--- a/junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
+++ b/junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
@@ -11,7 +11,10 @@ import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.openshift.OpenShifts;
 import cz.xtf.core.waiting.SimpleWaiter;
 import cz.xtf.junit5.config.JUnitConfig;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public class ProjectCreator implements TestExecutionListener {
     private static final OpenShift openShift = OpenShifts.master();
 
@@ -20,6 +23,24 @@ public class ProjectCreator implements TestExecutionListener {
         if (openShift.getProject() == null) {
             openShift.createProjectRequest();
             openShift.waiters().isProjectReady().waitFor();
+
+            try {
+                // Adding a label can be only done via 'namespace'. It cannot be set via 'project' API. Thus we do this
+                // separately. Also, to update namespace label, it's necessary to have 'patch resource "namespaces"'
+                // permission for current user and updated namespace, e.g. by having 'cluster-admin' role.
+                // Otherwise you can see:
+                // $ oc label namespace <name> "label1=foo"
+                // Error from server (Forbidden): namespaces "<name>" is forbidden: User "<user>" cannot patch resource "namespaces" in API group "" in the namespace "<name>"
+                OpenShifts.admin().namespaces().withName(openShift.getProject().getMetadata().getName()).edit().editMetadata()
+                        .addToLabels(OpenShift.XTF_MANAGED_LABEL, "true").endMetadata().done();
+            } catch (KubernetesClientException e) {
+                // We weren't able to assign a label to the new project. Let's just print warning since this information
+                // is not critical to the tests execution. Possible cause for this are insufficient permissions since
+                // some projects using XTF are executed on OCP instances without 'admin' accounts available.
+                log.warn("Couldn't assign label '" + OpenShift.XTF_MANAGED_LABEL + "' to the new project '"
+                        + openShift.getNamespace() + "'. Possible cause are insufficient permissions.");
+                log.debug(e.getMessage());
+            }
         }
         if (OpenShiftConfig.pullSecret() != null) {
             openShift.setupPullSecret(OpenShiftConfig.pullSecret());
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # XTF
-XTF is framework designed to easy up aspects of testing in OpenShift environment.
+XTF is a framework designed to ease up aspects of testing in OpenShift environment.
 
 
 ## Modules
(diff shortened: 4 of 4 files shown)


simkam:
LGTM, however, wouldn't it be better to add label in `OpenShift.createProjectRequest` and `OpenShift.recreateProject`? Any namespace created by XTF would then be labeled.

jstourac:
@simkam actually only in `OpenShift.createProjectRequest` since `recreateProject`will call it effectively.

Yeah, that was my first intention, although since I had to do this label setting in a different step after the project creation, I think that it is safer to wait until the project is prepared on the OpenShift instance first. Which I suppose is done via `openShift.waiters().isProjectReady().waitFor();`.

As such, putting this operation in the `createProjectRequest` method would change it's behavior a bit... Although, if this is not a problem, I can update.

simkam (approved):

core/src/main/java/cz/xtf/core/bm/BuildManager.java
```diff
@@ -16,6 +17,10 @@ public BuildManager(OpenShift openShift) {
         if (openShift.getProject(openShift.getNamespace()) == null) {
             openShift.createProjectRequest();
             openShift.waiters().isProjectReady().waitFor();
+            // Adding a label can be only done via 'namespace'. It cannot be set via 'project' API.
+            // Thus we do this separately. Also, updating namespaces seems to be restricted only to Admin users.
+            OpenShifts.admin().namespaces().withName(openShift.getNamespace()).edit().editMetadata().addToLabels(
```
mnovak1, line 22: Question I have is whether use of `admin` is required. Isn't `master` user enough?

jstourac: Using a regular `master` user led me into this error when I used this build in one of our project that utilizes this XTF:
```
Failure executing: PATCH at: https://api.nmsp-prne-wll3.eapqe.psi.redhat.com:6443/api/v1/namespaces/jstourac-namespace. Message: namespaces "jstourac-namespace" is forbidden: User "xpaasqe" cannot patch resource "namespaces" in API group "" in the namespace "jstourac-namespace". Received status: Status(apiVersion=v1, code=403, details=StatusDetails(causes=[], group=null, kind=namespaces, name=jstourac-namespace, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=namespaces "jstourac-namespace" is forbidden: User "xpaasqe" cannot patch resource "namespaces" in API group "" in the namespace "jstourac-namespace", metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Forbidden, status=Failure, additionalProperties={}).
org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
```

Also I can see similar error message, when I try manually via `oc` command:
```
$ oc label namespace test "label1=true"
Error from server (Forbidden): namespaces "test" is forbidden: User "xpaasqe" cannot patch resource "namespaces" in API group "" in the namespace "test"
```

Via `admin` user this operation succeeds:
```
$ oc label namespace test "label1=true" && oc delete project test
namespace/test labeled
```

Based on this, I suppose that `admin` user is required.

mnovak1: Ok, I'm not sure under which user namespace is created but if it's `admin` then I guess it's ok.

simkam: Both, `BuildManager` and `ProjectCreator` create a project as `master` user, not `admin`. I don't think it is important, `admin` is only needed to update labels on namespace.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -93,6 +93,7 @@
     private static volatile String routeSuffix;
 
     public static final String KEEP_LABEL = "xtf.cz/keep";
+    public static final String XTF_GENERATED_LABEL = "xtf.cz/generatedNamespace";
```
jstourac, line 96: I'm a bit thinking about this one thing now... I wonder whether the introduction of a completely new label is okay for you or whether usage of already existing `KEEP_LABEL` would be better. I have to say, that I really don't have exact idea what this existing label is supposed to be used for. As such, I'm not sure whether it is suitable also for my usage - labeling namespaces just for the purpose to identify they had been created by XTF.

simkam: `xtf.cz/keep` has a different purpose, it prevents resources from being cleaned by XTF as part of `OpenShift.clean()` method. The new label is ok for me, we might also generalize it a bit so it can be used also for other resources, `xtf.cz/managed` or something like that.

jstourac: I like that and I agree - although, with this we should put it on all resources so it really does based on its name... Are you okay if I modify it and add a comment there that at the moment we use this label only for the `namespaces` and that it can be extended later?

core/src/main/java/cz/xtf/core/bm/BuildManager.java
mchoma: There are cases we run xtf on clusters where we do not have admin rights. E.g. s390x testing. We run tests which does not need admin permissions there. 

But with this change introduced that will fail in such cases.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
```diff
@@ -20,6 +20,12 @@ public void testPlanExecutionStarted(TestPlan testPlan) {
         if (openShift.getProject() == null) {
             openShift.createProjectRequest();
             openShift.waiters().isProjectReady().waitFor();
+            // Adding a label can be only done via 'namespace'. It cannot be set via 'project' API.
+            // Thus we do this separately. Also, updating namespaces seems to be restricted only to Admin users.
+            OpenShifts.admin().namespaces().withName(
```
mchoma, line 25: Whenever using `.admin()` please specify in comment which permission we care about. This would be helpful for environments not having admin permissions.

In this case `patch resource "namespaces" in API group "" in the namespace "???"`

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
simkam: +1

mchoma (changes requested):
I have problem with using `.admin()`

core/src/main/java/cz/xtf/core/bm/BuildManager.java
simkam: good point, I think in that case we should only log warning that it isn't possible label namespace?

mchoma: If there really isn't way how to avoid xtf dependency on `admin()` rights than this is for sure better than fail.

So for environments where cluster_admin is not available it will stay basically same.

These cleaning job. I need to be admin I suppose to create them anyway? And if admin of shared cluster is willing to create such cleaning job, it can be still made like we have now e.g. `if (creator=xpaasuser && createdDate > 7days)` ?

I think this xtf optional dependency on permission `patch resource "namespaces" in API group "" in the namespace "???"` should be mentioned somewhere. I will let @mnovak1 identify what is the proper location. Is it README?

mchoma:
I have noticed this in other part of xtf to identify user resources. Would be something similar enough for clean job? 
```
	/**
	 * Retrieves secrets that aren't considered default. Secrets that are left out contain type starting with 'kubernetes.io/'.
	 *
	 * @return List of secrets that aren't considered default.
	 */
	public List<Secret> getUserSecrets() {
		return secrets().withLabelNotIn(OpenShift.KEEP_LABEL).list().getItems().stream()
				.filter(s -> !s.getType().startsWith("kubernetes.io/"))
				.collect(Collectors.toList());
	}
```

simkam:
The problem isn't to identify user resources but resources created by XTF and resources created manually by user.

core/src/main/java/cz/xtf/core/bm/BuildManager.java
```diff
@@ -16,6 +17,10 @@ public BuildManager(OpenShift openShift) {
...
+                    OpenShift.XTF_GENERATED_LABEL, "true").endMetadata().done();
```
istraka, line 23: It might be useful to have label either for xtf in generall (`xtf.cz/generatedNamespace`) but also for different type of namespaces - `xtf.cz/buildNamespace`, `xtf.cz/testNamespace` so that we can differ them easily.

Usecase: I want to easily delete all test namespace but no build namespace, since I still want to share some resources. Not sure if this scenario is valid, though.

Just thinking in long term...WDYT?

istraka (commented):
Just one optional thing about adding some labels.

core/src/main/java/cz/xtf/core/bm/BuildManager.java
jstourac: This looks like a good point. Although, I would probably keep it for the time we will really need it.

At the moment, I'll update this label to what Martin has proposed - `xtf.cz/managed`. :)

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
jstourac: I've updated in the comment what you mentioned.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
jstourac: Updated in latest `wip` commit

core/src/main/java/cz/xtf/core/bm/BuildManager.java
jstourac: Okay, I tried to put an if statement in latest `wip` commit so it wouldn't fail in case admin is not available. TBH - this doesn't seem to work yet, since e.g. if I add wrong admin password, then I get `java.lang.ExceptionInInitializerError`. Looks like current code is not able to cope with admin account not available? I'll check more into this later, but this change is to check whether you feel more comfortable with such solution.

jstourac: Also, I played a bit with the role binding yesterday and I found out that when I add a `cluster-admin` role-binding to the `xpaasqe` user on particular namespace, then he is then able to edit namespace lables... But I believe that we won't be able to provide such permissions, right? At least dynamically this would be quite tricky as such permission change has to be done by `admin` user anyway... Not sure whether we are able to have our OCP instance pre-configured somehow to incorporate that permission automatically.

mchoma: Yes, I am ok with if condition.

Our clusters is not problem, we could configure whatever we agree on. Problem is IBM cluster, which is maanged by Ops and they won't allow admin permission (somehow understandable). 

`ExceptionInInitializerError` interesting. Where is it from? when you use xpaasqe user in place of admin is it same?

core/src/main/java/cz/xtf/core/bm/BuildManager.java
```diff
@@ -16,6 +17,20 @@ public BuildManager(OpenShift openShift) {
         if (openShift.getProject(openShift.getNamespace()) == null) {
             openShift.createProjectRequest();
             openShift.waiters().isProjectReady().waitFor();
+
+            OpenShift adminOcp = OpenShifts.admin();
```
mchoma, line 21: when I use `xpaasqe` user in place of admin, it will be not null, right?

core/src/main/java/cz/xtf/core/bm/BuildManager.java
```diff
@@ -16,6 +17,20 @@ public BuildManager(OpenShift openShift) {
...
+            if (adminOcp != null) {
+                // Adding a label can be only done via 'namespace'. It cannot be set via 'project' API.
+                // Thus we do this separately. Also, to update namespace label, it's necessary to have 'cluster-admin'
+                // role binding for current user and updated namespace. Otherwise you can see:
+                // $ oc label namespace <name> "label1=foo"
+                // Error from server (Forbidden): namespaces "<name>" is forbidden: User "<user>" cannot patch resource "namespaces" in API group "" in the namespace "<name>"
+                OpenShifts.admin().namespaces().withName(openShift.getNamespace()).edit().editMetadata().addToLabels(
```
mchoma, line 28: Wouldn't be safest try this command, and catch exception and not fail on that?

core/src/main/java/cz/xtf/core/bm/BuildManager.java
jstourac: When I use wrong password (on purpose) for admin account :) I was just testing what happens. I'm not quite familiar with how this code is supposed to behave. As such, yes, this needs to be modified.

core/src/main/java/cz/xtf/core/bm/BuildManager.java
jstourac: Yeah

core/src/main/java/cz/xtf/core/bm/BuildManager.java
jstourac: Probably yes... it will be easier at least... my first intention was to extend the `OpenShifts` class to determine somehow whether we are able to gain admin account and privileges... but yeah, I guess I'm too lazy to do that now :)

jstourac:
Ready for re-review. I've updated and hopefully addressed all objections.

Regarding the label creation - I tried also with configuration with not sufficient admin privileges - no admin credentials, wrong admin credentials, used admin credentials with pre-filled values with regular user. In all cases the exception was thrown as expected and warning was logged. I hope that this is enough to test that and it won't break anything in our existing infrastructure.

core/src/main/java/cz/xtf/core/bm/BuildManager.java
```diff
@@ -16,6 +18,23 @@ public BuildManager(OpenShift openShift) {
         if (openShift.getProject(openShift.getNamespace()) == null) {
             openShift.createProjectRequest();
             openShift.waiters().isProjectReady().waitFor();
+
+            try {
+                // Adding a label can be only done via 'namespace'. It cannot be set via 'project' API.
+                // Thus we do this separately. Also, to update namespace label, it's necessary to have 'cluster-admin'
```
mchoma, line 24: > it's necessary to have 'cluster-admin'

This is not entirely true. We need just one permission. Can you see this permission in exception? Can we document that one permission here in comment as well. 

I would expect something similar
```
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://api.ocp44-001-cz7i.dynamic.xpaas:6443/api/v1/namespaces/73-q31f/pods. Message: pods is forbidden: User "xpaasqe" cannot list resource "pods" in API group "" in the namespace "73-q31f". Received status: Status(apiVersion=v1, code=403, details=StatusDetails(causes=[], group=null, kind=pods, name=null, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=pods is forbidden: User "xpaasqe" cannot list resource "pods" in API group "" in the namespace "73-q31f", metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Forbidden, status=Failure, additionalProperties={}).```

So permission is: `list resource "pods" in API group "" in the namespace`

mchoma: Otherwise looks good to me.

jstourac: Isn't `Error from server (Forbidden): namespaces "<name>" is forbidden: User "<user>" cannot patch resource "namespaces" in API group "" in the namespace "<name>"` what you mean?

mchoma: Yes it will be it. So permission will be `patch resource "namespaces"`. 

So can we reformulate sentence to something like:
`it's necessary to have permission to "patch resource namespaces", e.g. by having 'cluster-admin' role.`

mchoma: Oh, now I see. It is 2 lines below :)

mchoma (approved):

core/src/main/java/cz/xtf/core/bm/BuildManager.java
jstourac: Yes :) but I updated per your suggestion anyway since you are right that mentioning `cluster-admin` this way could be misleading. Hope it's good now.

simkam (approved):

mnovak1 (approved):
LGTM, merging.

PR #409, opened 2021-01-18
Do not attempt to delete kube-root-ca.crt config map

### LINKED ISSUE #406: Openshift.clean() gets stuck on OCP4.7
On OCP4.7, on every instance I've encountered. There is a configmap `kube-root-ca.crt` in every newly created namespace. Also this configmap can't be deleted. When I try to delete it, it won't be deleted (or possibly it is immediately created again?). 
This causes cleaning command `openshift.clean().waitFor();` to stuck, because configmaps in namespace will never be empty.
I'm not sure if this is bug or some new feature. I filed an [openshift issue](https://bugzilla.redhat.com/show_bug.cgi?id=1916688). Possibly this will require a change to cleanup method.

Crumby:
Fixes #406

Updates clean function of OpenShift.java class to be compatible with OCP4.7. Otherwise it gets stuck on attempt to remove `kube-root-ca.crt` file which is immediately recreated by OpenShift in every namespace.

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -1079,6 +1079,17 @@ public class OpenShift extends DefaultOpenShiftClient {
         return configMaps().list().getItems();
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


mnovak1 (approved):
LGTM

PR #422, opened 2021-03-26
WIP: Helm Client

spriadka:
- [x ] Pull Request does not include fixes for multiple issues/topics

This is a work in progress PR enabling support for Helm Client

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
+++ b/core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
@@ -1,0 +1,34 @@
+package cz.xtf.core.openshift;
+
+import java.util.Optional;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class HelmBinary {
+
+    private final String path;
+
+    @Getter
+    private String ocConfigPath;
+
+    public HelmBinary(String path) {
+        this.path = path;
+    }
+
+    public HelmBinary(String path, String ocConfigPath) {
+        this.path = path;
+        this.ocConfigPath = ocConfigPath;
+    }
+
+    public String execute(String... args) {
+        String[] prefix = Optional.ofNullable(ocConfigPath).map(ocConfig -> new String[] { path, "--kubeconfig", ocConfig })
+                .orElseGet(() -> new String[] { path });
+        return CLIUtils.executeCommand(ArrayUtils
+                .addAll(prefix, args));
+    }
+
+}
--- a/core/src/main/java/cz/xtf/core/openshift/HelmClients.java
+++ b/core/src/main/java/cz/xtf/core/openshift/HelmClients.java
@@ -1,0 +1,118 @@
+package cz.xtf.core.openshift;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Paths;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.SystemUtils;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+
+import cz.xtf.core.config.OpenShiftConfig;
+import cz.xtf.core.config.XTFConfig;
+import cz.xtf.core.http.Https;
+
+public class HelmClients {
+
+    private static final String HELM_CLIENTS_URL = "https://get.helm.sh";
+    private static final String HELM_FILE_NAME = "helm.tar.gz";
+
+    private static HelmBinary adminHelmBinary;
+    private static HelmBinary masterHelmBinary;
+
+    public static HelmBinary admin() {
+        if (adminHelmBinary == null) {
+            adminHelmBinary = getBinary(OpenShiftConfig.adminToken(),
+                    OpenShiftConfig.adminUsername(),
+                    OpenShiftConfig.adminPassword(),
+                    OpenShiftConfig.adminKubeconfig(),
+                    OpenShiftConfig.namespace());
+        }
+        return adminHelmBinary;
+    }
+
+    public static HelmBinary master() {
+        if (masterHelmBinary == null) {
+            masterHelmBinary = getBinary(OpenShiftConfig.masterToken(),
+                    OpenShiftConfig.masterUsername(),
+                    OpenShiftConfig.masterPassword(),
+                    OpenShiftConfig.masterKubeconfig(),
+                    OpenShiftConfig.namespace());
+        }
+        return masterHelmBinary;
+    }
+
+    private static HelmBinary getBinary(String token, String username, String password, String kubeconfig,
+            String namespace) {
+        OpenShiftBinary openShiftBinary = OpenShifts.getBinary(token, username, password, kubeconfig, namespace);
+        return new HelmBinary(downloadHelmBinary(), openShiftBinary.getOcConfigPath());
+
+    }
+
+    private static String downloadHelmBinary() {
+        String systemType = "";
+        String helmClientVersion = XTFConfig.get("helm.client.version", latestHelmVersion());
+        if (SystemUtils.IS_OS_MAC) {
+            systemType = "mac";
+        } else {
+            systemType = "linux";
+        }
+        //https://get.helm.sh/helm-v3.5.3-darwin-amd64.tar.gz
+        String architecture = System.getProperty("os.arch");
+        String helmClientUrl = String.format("%s/helm-%s-%s-%s.tar.gz", HELM_CLIENTS_URL, helmClientVersion, systemType,
+                architecture);
+        return downloadHelmBinaryInternal(helmClientVersion, helmClientUrl, true);
+    }
+
+    private static String latestHelmVersion() {
+        String gitApiResponse = Https.getContent("https://api.github.com/repos/helm/helm/releases/latest");
+        Gson gson = new Gson();
+        JsonObject jsonObject = gson.fromJson(gitApiResponse, JsonObject.class);
+        return jsonObject.get("tag_name").getAsString();
+    }
+
+    private static String downloadHelmBinaryInternal(final String version, final String helmClientUrl, final boolean trustAll) {
+        int code = Https.httpsGetCode(helmClientUrl);
+
+        if (code != 200) {
+            throw new IllegalStateException("Client binary for version " + version + " isn't available at " + helmClientUrl);
+        }
+
+        File workdir = helmBinaryFolder();
+
+        // Download and extract client
+        File helmTarFile = new File(workdir, HELM_FILE_NAME);
+        File helmFile = new File(workdir, "helm");
+
+        try {
+            URL requestUrl = new URL(helmClientUrl);
+
+            if (trustAll) {
+                Https.copyHttpsURLToFile(requestUrl, helmTarFile, 20_000, 300_000);
+            } else {
+                FileUtils.copyURLToFile(requestUrl, helmTarFile, 20_000, 300_000);
+            }
+
+            CLIUtils.executeCommand("tar", "-xf", helmTarFile.getPath(), "-C", workdir.getPath(), "--strip", "1");
+            FileUtils.deleteQuietly(helmTarFile);
+
+            return helmFile.getAbsolutePath();
+        } catch (IOException e) {
+            throw new IllegalStateException("Failed to download and extract oc binary from " + helmClientUrl, e);
+        }
+    }
+
+    private static File helmBinaryFolder() {
+        File workdir = new File(Paths.get("tmp/helm").toAbsolutePath().toString());
+        if (workdir.exists()) {
+            return workdir;
+        }
+        if (!workdir.mkdirs()) {
+            throw new IllegalStateException("Cannot mkdirs " + workdir);
+        }
+        return workdir;
+    }
+}
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
@@ -117,7 +117,7 @@ public class OpenShifts {
                 OpenShiftConfig.adminKubeconfig(), namespace);
     }
 
-    private static OpenShiftBinary getBinary(String token, String username, String password, String kubeconfig,
+    public static OpenShiftBinary getBinary(String token, String username, String password, String kubeconfig,
             String namespace) {
         String ocConfigPath = createUniqueOcConfigFolder().resolve("oc.config").toAbsolutePath().toString();
         OpenShiftBinary openShiftBinary;
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -66,6 +66,10 @@
 			<groupId>org.junit.jupiter</groupId>
 			<artifactId>junit-jupiter-engine</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.google.code.gson</groupId>
+			<artifactId>gson</artifactId>
+		</dependency>
 	</dependencies>
 	<build>
 		<plugins>
--- a/core/src/main/java/cz/xtf/core/openshift/CLIUtils.java
+++ b/core/src/main/java/cz/xtf/core/openshift/CLIUtils.java
@@ -1,0 +1,58 @@
+package cz.xtf.core.openshift;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.StringWriter;
+import java.util.Arrays;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import org.apache.commons.io.IOUtils;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class CLIUtils {
+
+    public static String executeCommand(String... args) {
+        ProcessBuilder pb = new ProcessBuilder(args);
+
+        pb.redirectOutput(ProcessBuilder.Redirect.PIPE);
+        pb.redirectError(ProcessBuilder.Redirect.PIPE);
+
+        try {
+            Process p = pb.start();
+
+            ExecutorService es = Executors.newFixedThreadPool(2);
+
+            Future<String> out = es.submit(() -> {
+                try (InputStream is = p.getInputStream(); StringWriter sw = new StringWriter()) {
+                    IOUtils.copy(is, sw);
+                    return sw.toString();
+                }
+            });
+
+            Future<String> err = es.submit(() -> {
+                try (InputStream is = p.getErrorStream(); StringWriter sw = new StringWriter()) {
+                    IOUtils.copy(is, sw);
+                    return sw.toString();
+                }
+            });
+
+            int result = p.waitFor();
+
+            if (result == 0) {
+                return out.get();
+            } else {
+                log.error("Failed while executing (code {}): {}", result, Arrays.toString(args));
+                log.error(err.get());
+            }
+        } catch (IOException | InterruptedException | ExecutionException e) {
+            log.error("Failed while executing: " + Arrays.toString(args), e);
+        }
+
+        return null;
+    }
+}
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java
@@ -1,15 +1,5 @@
 package cz.xtf.core.openshift;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.StringWriter;
-import java.util.Arrays;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.ArrayUtils;
 
 import lombok.Getter;
@@ -50,50 +40,9 @@ public class OpenShiftBinary {
     // Common method for any oc command call
     public String execute(String... args) {
         if (ocConfigPath == null) {
-            return executeCommand(ArrayUtils.addAll(new String[] { path }, args));
+            return CLIUtils.executeCommand(ArrayUtils.addAll(new String[] { path }, args));
         } else {
-            return executeCommand(ArrayUtils.addAll(new String[] { path, "--kubeconfig=" + ocConfigPath }, args));
+            return CLIUtils.executeCommand(ArrayUtils.addAll(new String[] { path, "--kubeconfig=" + ocConfigPath }, args));
         }
-    }
-
-    // Internal
-    private String executeCommand(String... args) {
-        ProcessBuilder pb = new ProcessBuilder(args);
-
-        pb.redirectOutput(ProcessBuilder.Redirect.PIPE);
-        pb.redirectError(ProcessBuilder.Redirect.PIPE);
-
-        try {
-            Process p = pb.start();
-
-            ExecutorService es = Executors.newFixedThreadPool(2);
-
-            Future<String> out = es.submit(() -> {
-                try (InputStream is = p.getInputStream(); StringWriter sw = new StringWriter()) {
-                    IOUtils.copy(is, sw);
-                    return sw.toString();
-                }
-            });
-
-            Future<String> err = es.submit(() -> {
-                try (InputStream is = p.getErrorStream(); StringWriter sw = new StringWriter()) {
-                    IOUtils.copy(is, sw);
-                    return sw.toString();
-                }
-            });
-
-            int result = p.waitFor();
-
-            if (result == 0) {
-                return out.get();
-            } else {
-                log.error("Failed while executing (code {}): {}", result, Arrays.toString(args));
-                log.error(err.get());
-            }
-        } catch (IOException | InterruptedException | ExecutionException e) {
-            log.error("Failed while executing: " + Arrays.toString(args), e);
-        }
-
-        return null;
     }
 }


core/src/main/java/cz/xtf/core/openshift/HelmClients.java
```diff
@@ -0,0 +1,108 @@
...
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.SystemUtils;
+
+import cz.xtf.core.config.OpenShiftConfig;
+import cz.xtf.core.http.Https;
+
+public class HelmClients {
+
+    private static final String HELM_CLIENTS_URL = "https://mirror.openshift.com/pub/openshift-v4/clients/helm/latest";
```
mchoma, line 16: I believe we would need `latest` to be parametrized. `latest` is good default, but like we talked before there could be situation like OCP EUS, which will need we choose different version.

spriadka: Fixed with e82acc779300f8a3c2130f7e031820ea6c71d9b5

mchoma:
Overal I think this is good. What is important to mention I haven't found any solid java API to be used, so seems we are really left to  handle helm by calling cli command. @spriadka did you considered java API as well?

mchoma (commented):
> Overal I think this is good. What is important to mention I haven't found any solid java API to be used, so seems we are really left to handle helm by calling cli command. @spriadka did you considered java API as well?

Although I came to this idea [1]. Tiller (helm server side) is protocol buffer and therefore could be generated with [2]. I think this is  worth to investigate.

[1] https://lairdnelson.wordpress.com/2017/06/14/helm-from-java/
[2] https://github.com/grpc/grpc-java

core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Collections;
+import java.util.Optional;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class HelmBinary {
```
mnovak1, line 16: At the first look, this looks like duplication of:
https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinary.java

together with `CLIUtils` class above.

Could be `OpenShiftBinary` used instead? Or delegated?

mchoma: At least here is `helmConfigPath` which differentiate. Of course some code can be shared but conceptually I think it is good idea to have separate classes for OpenShiftBinary (`oc` command) and HelmBinary (`helm` command)

mnovak1: Right, everything what is specific for `HelmBinary` can be in this class. This is about checking that code is not duplicated in `OpenShiftBinary`.

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
```diff
@@ -0,0 +1,109 @@
...
```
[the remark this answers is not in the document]
spriadka: Tiller support has been deprecated so we depend on wrapping the CLI calls. I've changed the URL for helm binary download from `mirror.openshift.com` take a look

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
```diff
@@ -0,0 +1,109 @@
...
+                    OpenShiftConfig.masterPassword(),
+                    OpenShiftConfig.masterKubeconfig(),
+                    OpenShiftConfig.namespace());
+        }
+        return masterHelmBinary;
+    }
+
+    private static HelmBinary getBinary(String token, String username, String password, String kubeconfig,
+            String namespace) {
+        OpenShiftBinary openShiftBinary = OpenShifts.getBinary(token, username, password, kubeconfig, namespace);
```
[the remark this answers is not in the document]
spriadka: `HelmBinary` is depending on `oc.config` that is being downloaded with the `oc.binary`.  The `OpenSjifts.getBinary()` calls downloads the binary, config and performs login on that config, so I found it convenient, as we'll stay in the same context as with OpenShift binary

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
```diff
@@ -0,0 +1,109 @@
...
+        return new HelmBinary(downloadHelmBinary(), openShiftBinary.getOcConfigPath());
```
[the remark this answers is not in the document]
spriadka: +1

core/src/main/java/cz/xtf/core/image/Product.java
```diff
@@ -29,6 +29,22 @@ public String templatesBranch() {
...
 
+    public String helmChartsRepo() {
+        return resolveDefaultingProperty("helm.charts.repo");
+    }
+
+    public String helmChartsBranch() {
+        return resolveDefaultingProperty("helm.charts.branch");
+    }
+
+    public String helmChartsLocation() {
```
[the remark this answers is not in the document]
spriadka: It's just parametrized directory within repository for helm charts (e.g https://github.com/wildfly/wildfly-charts/tree/main/charts). This just provides option to parameterized the location for charts.

core/src/main/java/cz/xtf/core/image/Product.java
```diff
@@ -29,6 +29,22 @@ public String templatesBranch() {
...
+        return resolveDefaultingProperty("helm.charts.location");
+    }
+
+    public String helmChartExamplesLocation() {
+        return resolveDefaultingProperty("helm.charts.examples.location");
```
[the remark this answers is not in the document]
spriadka: Simillarly to the charts parametrization, let's say we may have `examples-wildfly23, examples-wildfly24`. It really is just. about adding possibility. to adjust these properties to the repository layout

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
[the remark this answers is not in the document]
mchoma: > Tiller support has been deprecated so we depend on wrapping the CLI calls.

Sorry, I am failing to understand how you mean in this context 

> I've changed the URL for helm binary download from mirror.openshift.com take a look

I still think we should parametrize url and choose one default.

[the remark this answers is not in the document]
spriadka: So we'll end up with parametrized downloads URL and helm version? (Just asking to make it clear, the implementation is rather simple with this one)

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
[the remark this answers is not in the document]
spriadka: I guess we can have a workaround by passing env variables to helm binary. This may result in not requiring `oc.config` at all (https://helm.sh/docs/helm/helm/)

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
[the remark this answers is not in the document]
mchoma: Yes, I think it is enough. That will provide us flexibility to user to change what he need.

I think it is ok to default to `https://mirror.openshift.com/pub/openshift-v4/clients/helm/` and `latest` version. But still have way to download from `helm.sh` and specific version.

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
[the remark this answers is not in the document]
mchoma: I am not sure I follow. Maybe some chat or short call will help me. Are we trying to solve som real problem?

> I found it convenient, as we'll stay in the same context as with OpenShift binary

So is it about reusing `oc.config` path? Is solution for that just expose constant for `oc.config` and use it on both places? 

> I think we can do this with HELM_KUBETOKEN and HELM_KUBEASUSER variables

Or is it about sharing token? But we already have this parameters as input `String token, String username`

core/src/main/java/cz/xtf/core/image/Product.java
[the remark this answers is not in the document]
mchoma: So is this property required? If I do not specify `helm.charts.location` will it default to `charts` ?

core/src/main/java/cz/xtf/core/image/Product.java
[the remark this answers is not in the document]
mchoma: So is this property optional? If I do not specify `helm.charts.examples.location` will it default to `examples` ?

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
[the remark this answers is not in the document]
spriadka: We can 1) reuse `oc.config`, but in order to do that, we need to download it first. AFAIK only `OpenShiftBinary` uses this config and `OpenShift` is usually configured to use `OpenShiftConfig` internally (fabric8 API). `OpenShiftBinary` will download it with the binary itself, that's why I've decided to bundle `helm` binary with `oc` binary. 
2) If we don't want to re-use `oc.config` (or use it at all), we may not need to download it at all, and we can provide the environment variables to `ProcessBuilder`

AFAIK `oc.config` has `context` entries which point to user activity within namespace, so using `oc.config` can be fully replaced using 2)

core/src/main/java/cz/xtf/core/image/Product.java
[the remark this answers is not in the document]
spriadka: You're correct that the property is optional, however the defaulting behaviour is managed by test suites (or other software using XTF), in our use case, it will default to `charts`

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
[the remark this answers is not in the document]
mchoma: `reuse oc.config, but in order to do that, we need to download it first.` download from where? I thought when we talk about `oc.config` we talk about `~/.kube/config` local machine file filled with entries based on how `oc login` is used on that machine.

Is `oc.config` `~/.kube/config` ?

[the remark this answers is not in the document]
mchoma: Or in other words my understanding is we are taling here about 

```
$KUBECONFIG	set an alternative Kubernetes configuration file (default "~/.kube/config")
```
or
```
--kubeconfig string           path to the kubeconfig file
```
Described in https://helm.sh/docs/helm/helm/ . Am I missing something?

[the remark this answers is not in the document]
spriadka: I was talking about case of `OpenShifts.getBinary()` (https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShifts.java#L122) method, which will create empty `oc.config` in temporary directory, so the whole interaction of `OpenShitBinary` when running TS will not affect user's local `~/.kube/config`.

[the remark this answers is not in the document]
mchoma: Ok, so I am starting to understand it. `oc.config` is internal way of OpenShiftBinary how to keep authentication context. So you are basically delegating login logic to OpenShiftBinary. I understand that was easiest and most straighforward way.

Now I do not now if this is elegant or hack :) I tend to think it is hack :)

Conceptually I think HelmBinary and OpenShiftBinary should be independent.

I know there were cases where mystoriously in OpenShiftBinary authentication context was switching and I rather explicitely made .login() to ensure corect user is used with OpenShiftBinary. Probably it was because internal oc.config was changed.

In other words can it happen I `OpenShiftBinary.login()` with one user and I will switch user with `getHelmBinary()` call? I think that is dangerous.

But I still admit I do not see situation clear.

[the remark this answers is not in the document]
mchoma: @jbliznak this is discussion, why I am not sure if we want this in in this form.

core/src/main/java/cz/xtf/core/image/Product.java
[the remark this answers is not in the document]
mchoma: Ok, I agree default in xtf based on EAP knowledge could miss in other products. Not sure there is some convention among different products how helm repo is structured. I believe not.

Setting default in testsuite seems correct to me.

core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
```diff
@@ -117,7 +117,7 @@ public static OpenShiftBinary adminBinary(String namespace) {
                 OpenShiftConfig.adminKubeconfig(), namespace);
     }
 
-    private static OpenShiftBinary getBinary(String token, String username, String password, String kubeconfig,
```
mchoma, line 120: This is not necessary after deleting dependency between HelmBinary and OpenShiftBinary?

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
```diff
@@ -0,0 +1,107 @@
+package cz.xtf.core.openshift;
```
mchoma, line 1: What about putting helm specific things into own package, e.g. `cz.xtf.core.helm`?

After removing dependency between HelmBinary and OpenShiftBinary there is no reason to be part of `cz.xtf.core.openshift` ?

core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
```diff
@@ -0,0 +1,56 @@
+package cz.xtf.core.openshift;
```
mchoma, line 1: What about putting helm specific things into own package, e.g. cz.xtf.core.helm?

After removing dependency between HelmBinary and OpenShiftBinary there is no reason to be part of cz.xtf.core.openshift ?

core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
```diff
@@ -0,0 +1,56 @@
...
+        this.kubeToken = kubeToken;
+        this.namespace = namespace;
+    }
+
+    public String execute(String... args) {
+        Map<String, String> environmentVariables = new HashMap<>();
+        environmentVariables.put("HELM_CONFIG_HOME", helmConfigPath);
+        environmentVariables.put("HELM_KUBEAPISERVER", OpenShiftConfig.url());
+        environmentVariables.put("HELM_KUBEASUSER", kubeUsername);
+        environmentVariables.put("HELM_KUBETOKEN", kubeToken);
```
mchoma, line 51: Token can be empty. Does it work as expected in that case?

core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
```diff
@@ -0,0 +1,56 @@
...
+        this.kubeUsername = kubeUsername;
+        this.kubeToken = kubeToken;
+        this.namespace = namespace;
+    }
+
+    public String execute(String... args) {
+        Map<String, String> environmentVariables = new HashMap<>();
+        environmentVariables.put("HELM_CONFIG_HOME", helmConfigPath);
+        environmentVariables.put("HELM_KUBEAPISERVER", OpenShiftConfig.url());
+        environmentVariables.put("HELM_KUBEASUSER", kubeUsername);
```
mchoma, line 50: Username (`OpenShiftConfig.adminUsername()` or `OpenShiftConfig.masterUsername()`) can be empty. Does it work as expected in that case?

mchoma: I do not understand why password is missing here.

For OpenShift Client authentication is done either by token or by username/password. If token is used ofr authentication username is not necessary.

core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public HelmBinary(String path, String helmConfigPath, String kubeUsername, String kubeToken, String namespace) {
+        this.path = path;
+        this.helmConfigPath = helmConfigPath;
+        this.kubeUsername = kubeUsername;
+        this.kubeToken = kubeToken;
+        this.namespace = namespace;
+    }
+
+    public String execute(String... args) {
```
mchoma, line 46: So we are authenticating with each `execute` call? I am not saying that is bad, I just want to understand it.

For OpenShiftBinary, there was that idea of one login and subsequent calls in that authentication context.

I do not have preference right now. But did you compared both approaches and find authenticate on each call better?

core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
```diff
@@ -0,0 +1,56 @@
...
+        Map<String, String> environmentVariables = new HashMap<>();
+        environmentVariables.put("HELM_CONFIG_HOME", helmConfigPath);
```
mchoma, line 48: My feeling is we should use directly command parameters if this is possible instead of env variables.

Environmnet variables can be used to tweak behaviour from outside. but I do not have strong opinion.

core/src/main/java/cz/xtf/core/openshift/HelmClients.java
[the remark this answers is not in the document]
spriadka: @mchoma I agree on decoupling `OpenShiftBinary` and `HelmBinary`. I tried getting rid of `oc.config` and use env variables only (7e62dadf83252aa273b763b788cd25bdf5ad025e) but looks like, in addition to `HELM_KUBEASUSER`, `HELM_KUBETOKEN` and `HELM_KUBEAPISERVER`, we need to set `HELM_KUBECAFILE` env variable as well in order to establish secure connection with the K8S server.  This brings user rather complications into the usage of `HelmBinary` IMHO.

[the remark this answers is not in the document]
spriadka: `oc.config` is AFAIK temporary `kubeconfig` that in our case and our method calls, will be downloaded for each instance of `HelmBinary`. `OpenShiftBinary.login()` is being called (https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShifts.java#L129) everytime the `OpenShiftBinary` would is downloaded and every call of `OpenShifts.getBinary()` downloads new `OpenShiftBinary` instance so I don't think there would be any OpenShift context interference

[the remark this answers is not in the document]
mchoma: We have agreed on call. It seems it is not possible fully decouple OpenShiftBinary and HelmBinary.  The reason for that is HelmBinary does not have same capabilities [1] when it comes to authenticating to OpenShift as `oc`. Helm can authenticate with 
- kubeconfig - but we need to gain it with `oc`
- token - we can use it from test.properties, but still we do not have support for trustall and successfully create connection with SSL

[1] https://helm.sh/docs/helm/helm/#options

fabiobrz:
Hi @spriadka - Is there any chance to evaluate whether supporting Helm can be done without depending on (yet) another binary?
I haven't had any look at it but at a quick search I stumbled upon [this](https://github.com/microbean/microbean-helm), just as an example.

mchoma:
> Hi @spriadka - Is there any chance to evaluate whether supporting Helm can be done without depending on (yet) another binary? I haven't had any look at it but at a quick search I stumbled upon [this](https://github.com/microbean/microbean-helm), just as an example.

Hi @fabiobrz. We did consider java API, look [1] or [2]. We also found project you are mentioning, problem is it does not seem active anymore they stopped supporting version 2.x, wheares we need 3.x. We didn't found any fitting project back then. Seemed we would need to start writing something ourselves, which would be time consuming.

[1] https://github.com/xtf-cz/xtf/pull/422#issuecomment-815886464
[2] https://github.com/xtf-cz/xtf/pull/422#pullrequestreview-631318272

fabiobrz:
> We did consider java API, look [1] or [2]. We also found project you are mentioning, problem is it does not seem active anymore they stopped supporting version 2.x, wheares we need 3.x. We didn't found any fitting project back then. Seemed we would need to start writing something ourselves, which would be time consuming.

Thanks for your feedback @mchoma and sorry for not noticing the previous discussion about the topic.
I'll have a look around again, searching for alternatives (maybe a fork which carried the original project on?) because although ai perfectly understand the capacity issues, depending on binaries seems to me really the last resort choice.

mnovak1:
@spriadka Hi, could you update discussions in this PR, please? We would like to release new version of XTF with this merged. Thanks.

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManagerFactory.java
```diff
@@ -4,7 +4,7 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-enum OpenShiftBinaryManagerFactory {
+public enum OpenShiftBinaryManagerFactory {
```
simkam, line 7: why this change?

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
```diff
@@ -40,7 +40,7 @@ OpenShiftBinary adminBinary(String namespace) {
                 OpenShiftConfig.adminKubeconfig(), namespace);
     }
 
-    private OpenShiftBinary getBinary(String token, String username, String password, String kubeconfig,
+    public OpenShiftBinary getBinary(String token, String username, String password, String kubeconfig,
```
simkam, line 43: why this change?

core/src/main/java/cz/xtf/core/helm/HelmClients.java
```diff
@@ -0,0 +1,108 @@
...
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.SystemUtils;
+
+import cz.xtf.core.config.OpenShiftConfig;
+import cz.xtf.core.config.XTFConfig;
+import cz.xtf.core.http.Https;
+import cz.xtf.core.openshift.CLIUtils;
+
+public class HelmClients {
```
simkam, line 16: Could you please update this to follow the same pattern as oc binary.

public OpenShiftBinaryManagerFactory + package-private OpenShiftBinaryManager + path resolver(s)

it would be thread-safe and testable (also please add tests)

simkam (changes requested):

core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
spriadka: I have made quite some changes, would you have another look if this thread can be resolved?

core/src/main/java/cz/xtf/core/openshift/HelmBinary.java
mchoma: Ok

core/src/main/java/cz/xtf/core/openshift/OpenShifts.java
mchoma: I see this method was resolved from OpenShifts. Resolving.

README.md
```diff
@@ -247,6 +247,19 @@ JUnit5 module provides a number of extensions and listeners designed to easy up
...
+| Property name | Type | Description | Default value |
+----------------|------|-------------|---------------|
+| `xtf.helm.clients.url` | `String` | URL from which version specified by `xtf.helm.client.version` | `https://mirror.openshift.com/pub/openshift-v4/clients/helm` |
+| `xtf.helm.client.version` | `String` | Version of the Helm client to be downloaded (from `http://[xtf.clients.url]/[xtf.client.version`) | `latest` |
+| `xtf.helm.binary.path` | `String` | Path to existing Helm client binary. If absent, it will be downloaded using combination of `xtf.helm.clients.url` and `xtf.helm.client.version` parameters | |
+
 ## Releasing XTF
 Have a look to the [release documentation](RELEASE.md) to learn about the process that defines how to release XTF to 
 the community.
+JUnit5 module provides number of extensions and listeners designed to easy up OpenShift images test management. See [JUnit5](https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/waiting/SimpleWaiter.java) for more informations.
```
mchoma, line 263: Some copy paste error? I see same text above.

core/src/main/java/cz/xtf/core/config/HelmConfig.java
```diff
@@ -0,0 +1,26 @@
+package cz.xtf.core.config;
+
+import java.nio.file.Paths;
+
+public class HelmConfig {
+
+    public static final String HELM_BINARY_CACHE_PATH = "xtf.helm.binary.cache.path";
+    public static final String HELM_BINARY_CACHE_ENABLED = "xtf.helm.binary.cache.enabled";
+    public static final String HELM_BINARY_PATH = "xtf.helm.binary.path";
+    public static final String HELM_CLIENT_VERSION = "xtf.helm.client.version";
```
mchoma, line 10: So for consistency can you provide also method for `HELM_CLIENT_VERSION` and use that method in rest of MR?

core/src/main/java/cz/xtf/core/helm/HelmBinary.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public HelmBinary(String path, String helmConfigPath, String kubeUsername, String kubeToken, String namespace) {
+        this.path = path;
+        this.helmConfigPath = helmConfigPath;
+        this.kubeUsername = kubeUsername;
+        this.kubeToken = kubeToken;
+        this.namespace = namespace;
+    }
+
+    public String execute(String... args) {
```
mchoma, line 47: We talked about some fluent api would be nice for most frequent commands. That is out of scope of this MR?

mchoma:
I will be off one week. I do not have any important objections which could block this MR. I understand main exposed API for users are HelmBinary, HelmClients, HelmConfig, which seems robust enough to start with it. We can add next features as needed in future. From my perspective we can move on with this.

core/src/main/java/cz/xtf/core/helm/HelmClients.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public static HelmBinary adminBinary() {
+        return getBinary(OpenShiftConfig.adminToken(), OpenShiftConfig.adminUsername(), OpenShiftConfig.namespace());
+    }
+
+    public static HelmBinary masterBinary() {
+        return getBinary(OpenShiftConfig.masterToken(), OpenShiftConfig.masterUsername(), OpenShiftConfig.namespace());
+    }
+
+    private static HelmBinary getBinary(String token, String username, String namespace) {
```
simkam, line 15: this should be probably moved to HelmBinaryManager

core/src/main/java/cz/xtf/core/helm/HelmClients.java
```diff
@@ -0,0 +1,20 @@
+package cz.xtf.core.helm;
+
+import cz.xtf.core.config.OpenShiftConfig;
+
+public class HelmClients {
+
+    public static HelmBinary adminBinary() {
+        return getBinary(OpenShiftConfig.adminToken(), OpenShiftConfig.adminUsername(), OpenShiftConfig.namespace());
```
simkam, line 8: these methods should probably delegate to instantiable HelmBinaryManager

HelmBinaryManagerFactory.INSTANCE.getHelmBinaryManager().adminBinary()

core/src/main/java/cz/xtf/core/helm/ClusterVersionHelmBinaryPathResolver.java
```diff
@@ -0,0 +1,138 @@
...
+
+import cz.xtf.core.config.HelmConfig;
+import cz.xtf.core.config.XTFConfig;
+import cz.xtf.core.http.Https;
+import cz.xtf.core.openshift.ClusterVersionInfo;
+import cz.xtf.core.openshift.ClusterVersionInfoFactory;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class ClusterVersionHelmBinaryPathResolver implements HelmBinaryPathResolver {
```
simkam, line 23: package-private

core/src/main/java/cz/xtf/core/helm/ConfiguredPathHelmBinaryResolver.java
```diff
@@ -0,0 +1,10 @@
+package cz.xtf.core.helm;
+
+import cz.xtf.core.config.HelmConfig;
+
+public class ConfiguredPathHelmBinaryResolver implements HelmBinaryPathResolver {
```
simkam, line 5: package-private

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
```diff
@@ -40,7 +40,7 @@ OpenShiftBinary adminBinary(String namespace) {
                 OpenShiftConfig.adminKubeconfig(), namespace);
     }
 
-    private OpenShiftBinary getBinary(String token, String username, String password, String kubeconfig,
+    OpenShiftBinary getBinary(String token, String username, String password, String kubeconfig,
```
simkam, line 43: why this change?

simkam (changes requested):
tests don't pass for me locally so I didn't check them for now, but I might have messed rebase.

please rebase and squash commits where it makes sense

spriadka:
Thanks @simkam for resolving the threads, I got the CI passing (the problem was different child filename of the helm archives depending on the platform). I've also rebased and squashed commits

spriadka:
@simkam would you have another look at `getHelmClientUrlBasedOnConfiguredHelmVersion` and the `resolve` methods? Please resolve the threads if you're satisfied with the proposed solutions

core/src/main/java/cz/xtf/core/helm/ClusterVersionHelmBinaryPathResolver.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.commons.lang3.SystemUtils;
+
+import cz.xtf.core.config.HelmConfig;
+import cz.xtf.core.http.Https;
+import cz.xtf.core.openshift.ClusterVersionInfo;
+import cz.xtf.core.openshift.ClusterVersionInfoFactory;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+class ClusterVersionHelmBinaryPathResolver implements HelmBinaryPathResolver {
```
simkam, line 23: can you please rename this and related tests? Perhaps something like `ConfiguredVersionHelmBinaryPathResolver`

core/src/main/java/cz/xtf/core/helm/ClusterVersionHelmBinaryPathResolver.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private static final String OCP_4_HELM_BINARY_DOWNLOAD_URL = "https://mirror.openshift.com/pub/openshift-v4/clients/helm";
+
+    @Override
+    public String resolve() {
+        final ClusterVersionInfo clusterVersionInfo = ClusterVersionInfoFactory.INSTANCE.getClusterVersionInfo();
+        if (!clusterVersionInfo.getOpenshiftVersion().startsWith("4")) {
+            log.info(
```
simkam, line 31:
```suggestion
            log.warn(
```

simkam (changes requested):
Thanks, last two minor comments and we can merge this.

simkam (approved):
LGTM, but please squash commits

PR #436, opened 2021-05-18
Allow change to PostgreSQL and MySQL readiness and liveness probe settings

vsalbaba:
Allow XTF library user to change probe settings by overriding `getProbeSettings` in his classes
Do not change behaviour and probe settings for Derby, MySQL, and PostgreSQL

What I did:
* Pulled duplicate code for liveness and readiness up into `AbstractSQLDatabase.java`
* Extracted MySQL and PostgreSQL probe settings from `configureContainer` to `ProbeSettings.java`

XTF library user will:
* derive his own class, for example `GalleonConfiguredDatasourcePostgreSQL extends PostgreSQL`
* optionally override `getProbeSettings`, using `super.getProbeSettings()` and updating them:
```java
public class GalleonConfiguredDatasourcePostgreSQL extends PostgreSQL {
// ...
	@Override
	protected ProbeSettings getProbeSettings() {
		ProbeSettings settings = super.getProbeSettings();
		settings.setReadinessInitialDelaySeconds(10);
		return settings;
	}
// ...
}
```

### DIFF
--- a/builder/src/main/java/cz/xtf/builder/db/AbstractSQLDatabase.java
+++ b/builder/src/main/java/cz/xtf/builder/db/AbstractSQLDatabase.java
@@ -7,6 +7,7 @@ import java.util.function.Consumer;
 
 import org.apache.commons.io.IOUtils;
 
+import cz.xtf.builder.builders.pod.ContainerBuilder;
 import cz.xtf.builder.builders.pod.PersistentVolumeClaim;
 
 public abstract class AbstractSQLDatabase extends AbstractDatabase implements SQLExecutor {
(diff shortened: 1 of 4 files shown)


builder/src/main/java/cz/xtf/builder/db/AbstractSQLDatabase.java
```diff
@@ -77,5 +78,32 @@ public SQLExecutor getSQLExecutor(String hostname, int port) {
                 getPassword());
     }
 
+    @Override
+    protected void configureContainer(ContainerBuilder containerBuilder) {
```
mchoma, line 82: So the main question is. Is this change backward compatible?

This is `protected`, so I assume you have updated all occurences in xtf. But in theory `AbstractSQLDatabase` can be used in any of testsuites. But in that case its overrided configureContainer should be used. So this seems to me fully  backward compatible.

Am I missing something?

mchoma:
I like this approach. Alternativlly we were talking about introducing `xtf.*` properties to have way for users to override this behaviour. But this api change seems to me more elegant and robust.

mchoma (approved):

mchoma:
@mnovak1 please could you have a look. We would like to leverage this in EAP 7.4 testing.

simkam (approved):

builder/src/main/java/cz/xtf/builder/db/AbstractSQLDatabase.java
vsalbaba: There are 3 classes inheriting from `AbstractSQLDatabase` and each was overriding `configureContainer` from `AbstractDatabase` in xtf - 
* `PostgreSQL` and `MySQL` are now using this one, both methods did the same thing with different params
* `Derby` now instead of overriding `AbstractDatabase#configureContainer` overrides `AbstractSQLDatabase#configureContainer`, which will not change behaviour
* `Derby` now has inherits `getProbeSettings`, but it will raise an exception if called. `Derby` does not call it and I cant imagine new class inherited from `PostgreSql` which would call `configureContainer` without its own `ProbeSettings`. It has to specify command to check readiness with.

I'm pretty sure this is fully backwards compatible.

vsalbaba:
Attached log from running test-eap com.redhat.xpaas.eap.db.MySQLTest and com.redhat.xpaas.eap.db.PostgreSQLTest with these changes from local repository - succeded.
[xtf-test-eap-openshift test.log](https://github.com/xtf-cz/xtf/files/6501442/xtf-test-eap-openshift.test.log)
[xtf-eap-openshit-postgresql-test.log](https://github.com/xtf-cz/xtf/files/6501440/xtf-eap-openshit-postgresql-test.log)

mnovak1 (approved):

mnovak1:
Merging, thanks

PR #437, opened 2021-05-19
Do not require pgp key when deploying XTF snapshot

mnovak1:

### DIFF
--- a/.github/workflows/xtf-maven-snapshots.yaml
+++ b/.github/workflows/xtf-maven-snapshots.yaml
@@ -2,12 +2,12 @@ name: XTF Maven Release
 
 on:
   push:
-    tags:
+    branches:
       - '**'
 jobs:
 
   test:
-    name: Deploy to JBoss Release Maven repo
+    name: Deploy to JBoss Snapshots Maven repo
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/xtf-maven-snapshots.yaml
```diff
@@ -0,0 +1,36 @@
+name: XTF Maven Release
+
+on:
```
mchoma, line 3: Are we sure we want this? Aren't we discussing this topic in chat that we dont want tu create snapshots from all forks?

mnovak1: I think it's good feature to allow to deploy snapshot from forks as it allow users to test their features in consuming test suites. This PR is just to simplify it that PGP key is not required for snapshots. The other modification will be handled in separate PR.

mchoma: As I understand this will `deploy snapshot from forks`. I am not fan of this. We can talk on Monday meeting how you mean and what other think.

mnovak1: Sure, we can incorporate other changes we've discussed as well

Issue #438, opened 2021-05-24
OpenShiftWaiters is using hardcoded timeouts


fabiobrz:
Looking at e.g. https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java#L61 (and other locations) an hardcoded value of 10 mins is used for the returned `Waiter`. 

Is there a reason for doing so instead of using e.g.: `WaitingConfig.timeout()` as it's done at https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java#L128 ?

We've found ourselves forced to implement such a tiny variation locally and this could be fine but I am wondering what the added value of the original (with hardcoded timeout) is in such a case.

If there is consensus on fixing the reported methods I can provide a PR.

CC  @mchoma @mnovak1 @simkam

mchoma:
Your suggestion make sense to me.

PR #439, opened 2021-05-26
Replacing hardcoded 10 mins waiting with WaitingConfig.timeout for OpenShiftWaiters#hasBuildCompleted

fabiobrz:
Using configuration values rather than some hardcoded timeouts in OpenShiftWaiters.
 
Fix https://github.com/xtf-cz/xtf/issues/438

FYI @mchoma

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
@@ -45,8 +45,9 @@ public class OpenShiftWaiters {
     }
 
     /**
-     * Creates waiter for latest build completion with preconfigured timeout 10 minutes,
-     * 5 seconds interval check and both logging points.
+     * Creates waiter for latest build completion with a build timeout which is preconfigured to be up to 10 minutes
+     * (but can also be customized by setting {@code xtf.waiting.build.timeout}), 5 seconds interval check and both
+     * logging points.
      *
      * @param buildConfigName build name to wait upon
      * @return Waiter instance
@@ -58,15 +59,17 @@ public class OpenShiftWaiters {
                 .orElse(null);
         String reason = "Waiting for completion of latest build " + buildConfigName;
 
-        return new SupplierWaiter<>(supplier, "Complete"::equals, "Failed"::equals, TimeUnit.MINUTES, 10, reason)
-                .logPoint(Waiter.LogPoint.BOTH)
-                .failFast(failFast)
-                .interval(5_000);
+        return new SupplierWaiter<>(supplier, "Complete"::equals, "Failed"::equals, TimeUnit.MINUTES,
+                WaitingConfig.buildTimeout(), reason)
+                        .logPoint(Waiter.LogPoint.BOTH)
+                        .failFast(failFast)
+                        .interval(5_000);
     }
 
     /**
-     * Creates waiter for build completion with preconfigured timeout 10 minutes,
-     * 5 seconds interval check and both logging points.
+     * Creates waiter for build completion with a build timeout which is preconfigured to be up to 10 minutes (but can
+     * also be customized by setting {@code xtf.waiting.build.timeout}), 5 seconds interval check and both logging
+     * points.
      *
      * @param build to be waited upon.
      * @return Waiter instance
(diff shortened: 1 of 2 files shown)


mchoma:
This is aligned with what we have discussed. I am approving it.

mchoma (approved):

core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
```diff
@@ -56,29 +56,31 @@ public Waiter hasBuildCompleted(String buildConfigName) {
...
                 .map(openShift::getBuild).map(Build::getStatus).map(BuildStatus::getPhase)
                 .orElse(null);
-        String reason = "Waiting for completion of latest build " + buildConfigName;
+        final String reason = "Waiting for completion of latest build " + buildConfigName;
 
-        return new SupplierWaiter<>(supplier, "Complete"::equals, "Failed"::equals, TimeUnit.MINUTES, 10, reason)
-                .logPoint(Waiter.LogPoint.BOTH)
-                .failFast(failFast)
-                .interval(5_000);
+        return new SupplierWaiter<>(supplier, "Complete"::equals, "Failed"::equals, TimeUnit.MINUTES, WaitingConfig.timeout(),
```
mnovak1, line 61: Current default timeout is 180s (3 min) which is much less then 10min. In case we increase `xtf.waiting.timeout` then other timeouts (not just for the build) will take much longer. We might consinder to create timeout just for the build completion with 10 min default and allow to reconfigure. 

cc-ing @mchoma

fabiobrz: Thanks @mnovak1

> Current default timeout is 180s (3 min) which is much less than 10min.

Ok, but that is just the default (or better to say a fallback value for it).But I wouldn't change it. Users should be aware of what methods use timeout and provide their default accordingly, if they want something more than 3 mins.
WDYT?

mchoma: @mnovak1 I agree with you. I was not sure if we are ok with introducing new properties. But if you are ok we can do it like that. We have discussed that with @fabiobrz and he will change accordingly - introduce new property `xtf.waiting.build.timeout` especially for this usage with default 10 minutes.

fabiobrz: Hi @mnovak1 - As a second thought and after some discussion with @mchoma I think get your point is that the library itself relies on the hardcoded business logic, i.e. internal calls assume that this method will have a timeout of 10 mins, and the default fallback of 3 mins could harm those calls. 
IMHO this is actually something that could be improved because the whole point of this PR is that those calls timeouts should be changed into something depending on configuration.

So I'll move on by following your suggestion and not touch this method at all, while I'll create a new one. 
Maybe the existing method should be named somehow differently to include the "hardcoded Timeout" concept.

mnovak1: Purpose is to allow to set timeout for builds. Now the timeout was hardcoded to 10 min which is safe value for most of the builds. By changing it to `xtf.waiting.timeout` which is 3 min by default, this timeout will be considerably shortened and it might cause timeout failures in consuming test suites as 3min is really not much time for builds.  

Thus I think we need to add another property to configure it.

mchoma: > I think get your point is that the library itself relies on the hardcoded business logic, i.e. internal calls assume that this method will have a timeout of 10 mins

IIUC No, I think this is not happening.

I would vote now for introducing `xtf.waiting.build.timeout` with default 10 minutes and use it in this particular case.

New method could help to avoid need to introduce new property, but that would mean all callers should change to apply this change.

fabiobrz: Ok, thanks @mnovak1 and @mchoma for your feedback, I am going to update the PR by following your suggestions.

fabiobrz:
HI @mnovak1 - I tried to address your comments. Let me know WDYT.

mchoma:
> HI @mnovak1 - I tried to address your comments. Let me know WDYT.

LGTM

mnovak1 (approved):
LGTM. Merging

PR #454, opened 2021-08-25
Add feature cache oc binary in local folder.

mcarlett:

### DIFF
--- a/core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
+++ b/core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
@@ -1,5 +1,7 @@
 package cz.xtf.core.config;
 
+import java.nio.file.Paths;
+
 public final class OpenShiftConfig {
     public static final String OPENSHIFT_URL = "xtf.openshift.url";
     public static final String OPENSHIFT_TOKEN = "xtf.openshift.token";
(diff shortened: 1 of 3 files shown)


simkam:
so this is supposed to cache binary between multiple executions? Currently binary is already download only once during execution.

mcarlett:
@simkam yes , this feature can be useful if we use the library in multiple automated test, besides of multiple manual executions in the same host (replace n downloads of the 100+MB with just 1 download)

core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
```diff
@@ -6,6 +6,9 @@
     public static final String OPENSHIFT_VERSION = "xtf.openshift.version";
     public static final String OPENSHIFT_NAMESPACE = "xtf.openshift.namespace";
     public static final String OPENSHIFT_BINARY_PATH = "xtf.openshift.binary.path";
+    public static final String OPENSHIFT_BINARY_CACHE_ENABLED = "xtf.openshift.binary.cache.enabled";
+    public static final String OPENSHIFT_BINARY_CACHE_PATH = "xtf.openshift.binary.cache.path";
+    public static final String OPENSHIFT_BINARY_CACHE_PATH_DEFAULT_VALUE = "/tmp/xtf/oc-cache";
```
mnovak1, line 11: I think there could be used more OS agnostic approach and store `oc` in folder like: 
```Files.createTempDirectory(...);```

mnovak1 (commented):
I have suggestion for one improvement otherwise LGTM.

core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
mcarlett: sure good idea, thanks @mnovak1 I will apply this change

simkam: Wouldn't it effectively disable caching with default value? A new temporary directory would be created with each execution, no?

mcarlett: @simkam I update the PR, I cannot create a new temp directory every time, I just used java.io.tmpdir property to retrieve the temp root dir, and I used Paths.get(..) to concatenate root temp and folder

simkam (approved):

mnovak1 (approved):

PR #475, opened 2022-01-26
Adding the Service Logs Streaming feature

### LINKED ISSUE #474: Implement a solution for streaming pod logs
This requirement is the result of experiments made in order to stream pod logs directly to the users terminal, without requiring for them to access the cluster.

FYI  @mchoma

fabiobrz:
Fix #474 

CC @mchoma, @simkam, @tommaso-borgato

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,91 @@ xtf.foo.v2.version=1.0.3
 
 Retrieving an instance with this metadata: `Produts.resolve("product");`
 
+#### Service Logs Streaming (SLS)
+This feature allows for you to stream the services output while the test is running; this way you can see immediately 
+what is happening inside the cluster.
+This is of great help when debugging provisioning, specifically on Cloud environments, which instead would require for
+you to access your Pods.
+
+##### Kubernetes/OpenShift implementation
+The SLS OpenShift platform implementation relies upon the following fabric8 Kubernetes Client API features:
+
+- Watching Kubernetes events (see
+  [PodWatchEquivalent.java](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/kubectl/equivalents/PodWatchEquivalent.java))
+- Watching Pod logs (see
+  [PodLogExample.java](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/PodLogExample.java))
+
+The expected behavior is to stream the output of all the containers that are started or terminated in the selected 
+namespaces.
+
+##### Usage
+The SLS feature can be configured and enabled either via annotations or via properties.
+This behavior is provided by 
+[the ServiceLogsStreamingRunner JUnit 5 extension](./junit5/src/main/java/cz/xtf/junit5/extensions/ServiceLogsStreamingRunner.java).
+There are two different ways for enabling the SLS functionality, which are summarized in the following sections, please 
+refer to the [JUnit 5 submodule documentation](./junit5/README.md) in order to read about the extension implementation 
+details.
+
+###### The `@ServiceLogsStreaming` annotation (Developer perspective)
+Usage is as simple as annotating your test with `@ServiceLogsStreaming` e.g.:
+
+```java
+@ServiceLogsStreaming
+@Slf4j
+public class HelloWorldTest {
+  // ...
+}
+```
+
+###### The `xtf.log.streaming.enabled` and `xtf.log.streaming.config` property (Developer/Automation perspective)
+You can enable the SLS feature by setting the `xtf.log.streaming.enabled` property so that it would apply to 
+all the test classes being executed. 
+
+Conversely, if the above property is not set, you can set the `xtf.log.streaming.config` 
+property in order to provide multiple SLS configurations which could map to different test classes. 
+
+The `xtf.log.streaming.config` property value is expected to be a _comma_ (`,`) separated list of _configuration items_,
+each one formatted as a _semi-colon_ (`;`) separated list of _name_ and _value_ pairs for the above mentioned
+attributes, where the name/value separator is expected to be the _equals_ char (`=`). 
+A single _configuration item_ represents a valid source of configuration for a single SLS activation and exposes the 
+following information:
+
+* _**target**_: a regular expression which allows for the testing engine to check whether the current context test class
+    name matches the Service Logs Streaming configuration - **REQUIRED**
+
+* _**filter**_: a string representing a _regex_ to filter out the resources which the Service Logs Streaming activation
+    should be monitoring - **OPTIONAL**
+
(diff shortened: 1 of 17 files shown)


mchoma:
I am looking forward to use this feature :)

mchoma:
I could not resist to use it immediatelly :)  and it works!!! Just colouring does not work in my case, but could be just me.

fabiobrz:
> I could not resist to use it immediatelly :) and it works!!! Just colouring does not work in my case, but could be just me.

Thanks for your feedback @mchoma ! I'd like to check about the coloring thing, how did you enable the feature? e.g.: did you pass `xtf.sls.enabled=true`? This is for me to understand whether logs are sent to STDOUT or to a file.

Actually this wouldn't change a thing because I've been using `cat` to see a generated log file and I could see colors there as well, but I'd focus on your use case to double check.

mnovak1:
@fabiobrz I like this idea. This is something we miss and will be useful for everyone. Thanks for raising PR. I'll review it as soon as possible :-)

junit5/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
```diff
@@ -0,0 +1 @@
+cz.xtf.junit5.extensions.ServiceLogsStreamingRunner
```
simkam, line 1: other extensions aren't enabled by default in services, why this should be an exception?

simkam (changes requested):
Could you please review the visibility of classes? It looks like all are public but are only used by extension within the package, perhaps they can be changed to package-private? Let's expose as minimum as possible.

junit5/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
fabiobrz: Thanks @simkam - This was the solution that came to my mind when I though about how to enable the feature. 
Per se, it had a requirement allowing for it to be started either via annotation (on individual classes for developer UX) or via system properties (automation use case).
I thought a Before/After-AllCallback was the best way to hook to the workflow, hence the extension.
And I thought the best way for that kind of extension to be not bound to annotations was to be enabled by default and then to check for system properties or annotations in order to eventually enable the feature.

Basically what is enabled by default is a _controller_, implemented by a JUnit extension, that enable the feature (part of the core) based on some conditions but please let me know whether you'd suggest something different for that part.

fabiobrz:
> Could you please review the visibility of classes? It looks like all are public but are only used by extension within the package, perhaps they can be changed to package-private? Let's expose as minimum as possible.

Will do, thanks @simkam

junit5/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
simkam: I understand, but it is different from other extensions do here in XTF. With other extensions, users need explicitly load them in their projects, no matter whether the feature is disabled or enabled.

fabiobrz: Sure, I see. I'll think about the changes and hopefully come up with something soon. Thanks for your feedback.

mnovak1:
@fabiobrz For me there was no issue to enable the feature and have colored text on standard output. I wanted to try how it will behave when forwarded to a file like `-Dxtf.sls.config="target=.*Test;output=/home/mnovak/sls-logs.txt"` but no `/home/mnovak/sls-logs.txt` file was created. It's still logged to standard output. Is there anything else needed to configure?

fabiobrz:
> @fabiobrz For me there was no issue to enable the feature and have colored text on standard output. I wanted to try how it will behave when forwarded to a file like `-Dxtf.sls.config="target=.*Test;output=/home/mnovak/sls-logs.txt"` but no `/home/mnovak/sls-logs.txt` file was created. It's still logged to standard output. Is there anything else needed to configure?

Hi @mnovak1 - Actually `output` should point to a directory which is meant to be the base path in which files for each executed test class will be created.
There are (at least) a couple of issues, though:
1. the implementation should create such path if it does not exist (?)
2. it should be better documented that the expected path is about a directory rather than a file.

Could you please take such information into account and try again? I had my files generated and would appreciate a confirmation from you.
I'll fix what above as soon as I push something to implement the changes in the feature activation logic that have been previously requested.

simkam:
wrt feature activation, I think that only removing it from `services` and documenting it should be enough. Just to keep it consistent with other extensions.

fabiobrz:
> wrt feature activation, I think that only removing it from `services` and documenting it should be enough. Just to keep it consistent with other extensions.

Thanks @simkam - so basically according to your suggestion, the users should:

- register the extension, e.g. adding the FQN to `services` locally  
- use the system property or annotation to enable the feature when running tests

Correct?

simkam:
Yes, exactly

mnovak1:
> Hi @mnovak1 - Actually `output` should point to a directory which is meant to be the base path in which files for each executed test class will be created. There are (at least) a couple of issues, though:
> 
> 1. the implementation should create such path if it does not exist (?)
> 2. it should be better documented that the expected path is about a directory rather than a file.
> 
> Could you please take such information into account and try again? I had my files generated and would appreciate a confirmation from you. I'll fix what above as soon as I push something to implement the changes in the feature activation logic that have been previously requested.

Thanks, now it works. +1 for both points.

simkam:
> About this, @simkam - actually the extension lives in a different module and package as such, hence by a quick check, only PodLogsWatcher could be package-private. Am I missing something about this specific request?

No, sorry, I didn't notice extension is in different package.

junit5/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
fabiobrz: Extension registration removed, resolving this thread.

fabiobrz:
Hi @simkam - I pushed some changes in order to address your comments. Feel free to let me know whether you've got any additional feedback.

simkam (approved):

simkam:
LGTM

README.md
```diff
@@ -157,6 +157,91 @@ xtf.foo.v2.version=1.0.3
...
+```java
+@ServiceLogsStreaming
+@Slf4j
+public class HelloWorldTest {
+  // ...
+}
+```
+
+###### The `xtf.sls.enabled` and `xtf.sls.config` property (Developer/Automation perspective)
+You can enable the SLS feature by setting the `xtf.sls.enabled` property so that it would apply to
```
mnovak1, line 197: Just a suggestion. `sls` seems to hard to remember (at least to me :-) ) Could it be renamed to `xtf.log.streaming.enabled` ? wdyt?

fabiobrz: Thanks @mnovak1, will do.

Issue #325, opened 2019-10-29
XTF is missing test coverage


rsvoboda:
XTF is missing test coverage

I haven't find any tests. Please prove me wrong.

dsimansk:
That's absolutely correct, there're none. Can you share any expectation for minimal acceptable test coverage?

rsvoboda:
I expect test coverage for key functionalities so people have guidance for the usage

mnovak1:
Missing test coverage is one of the biggest painpoints of XTF framework. Currently there is basically no way how to provide reasonable guarantees that backward compatibility is not broken and there are no regressions which makes upgrades to new XTF version expensive as time is burned on fixing issues. 

As a start test should be provided to test API of main classes:
`OpenShifts`
`OpenShift`
`Waiters`
`Waiter`
`Image`
`BuildManager`
`Https`

as those seem to be heavily used.

mnovak1:
Any ideas how the test suite could be run against real Openshift instance as part of PR review process?

Looks like that kubernetes-client/openshift-client project is mocking Openshift server. For example see test [1]. It might be tricky to use it for XTF as it does not have to reflect changes in API model in new Openshift versions or reflect precisely current Openshift API.

[1] https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/OpenshiftRoleTest.java

mnovak1:
I've started to evaluate possibility to use Openshift CI (https://steps.svc.ci.openshift.org/help) for verification of PRs during dev review. admins/reviewers would have rights to trigger XTF TS against real OCP 4 instance. Local TS runs will need its own OCP 4 instance. From my pov it's better than mocking Openshift as it reflects changes in OCP 4 API. Drawback is that TS will take longer to run (in comparison to mocking) and there is no OCP 3 instance in Openshift CI for testing of backward compatibility. But it's better than current status and currently best solution I've found. 

Any objections to this way?

fabiobrz:
Here I am @mnovak1 - after closing #476. 
I've gone through the comments and noticed that https://steps.svc.ci.openshift.org/help is dead.
IMHO we need to find some more reliable solution in case we don't want to use the mock.

BTW - using the mock is a quick and easy way to start and build the internal TS.
Conversely, thinking about testing against an actual OCP backend would involve more steps and complexity (BTW this is demonstrated that the URL is dead after two years since the last comment).

[...]

PR #477, opened 2022-02-16
Baseline for OpenShift tests using a mocked OpenShift server

fabiobrz:
Following up ion the discussion in https://github.com/xtf-cz/xtf/issues/325#issuecomment-1041556773, this (Draft) PR is a proposal for how to add some smoke tests that would check the components provided by XTF against a mocked OpenShift server.

The changes here are testing CRUD operations provided by the `OpenShift` class on a copule of resources, namely `ConfigMap` and `Deployment`, in order to prove the concept.

The PR itself could be extended conveniently.
@simkam, @mnovak1 WDYT? Could this way work for you as well?

### DIFF
--- a/builder/pom.xml
+++ b/builder/pom.xml
@@ -27,6 +27,12 @@
 			<artifactId>lombok</artifactId>
 			<scope>provided</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>io.fabric8</groupId>
+			<artifactId>openshift-server-mock</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 </project>
--- a/builder/src/test/java/cz/xtf/builder/openshift/mocked/smoke/BasicOpenShiftTest.java
+++ b/builder/src/test/java/cz/xtf/builder/openshift/mocked/smoke/BasicOpenShiftTest.java
@@ -1,0 +1,129 @@
+package cz.xtf.builder.openshift.mocked.smoke;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Assert;
+import org.junit.jupiter.api.Test;
+
+import cz.xtf.builder.builders.ConfigMapBuilder;
+import cz.xtf.builder.builders.DeploymentConfigBuilder;
+import cz.xtf.core.openshift.OpenShift;
+import io.fabric8.kubernetes.api.model.ConfigMap;
+import io.fabric8.kubernetes.api.model.PodSpec;
+import io.fabric8.openshift.api.model.DeploymentConfig;
+import io.fabric8.openshift.client.OpenShiftClient;
+import io.fabric8.openshift.client.server.mock.OpenShiftServer;
+
+public class BasicOpenShiftTest {
+
+    private static final String TEST_RESOURCE_LABEL_NAME_APP = "app";
+    private static final String TEST_RESOURCE_LABEL_VALUE_APP = "xtf-core-test-openshift-mocked-smoke";
+    private final static Map<String, String> TEST_RESOURCE_LABELS = Collections.singletonMap(
+            TEST_RESOURCE_LABEL_NAME_APP, TEST_RESOURCE_LABEL_VALUE_APP);
+
+    private static final String TEST_CONFIGMAP_NAME = "test-configmap";
+    private static final String TEST_DEPLOYMENT_NAME = "test-deployment";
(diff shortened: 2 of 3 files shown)


core/src/test/java/cz/xtf/core/openshift/mocked/smoke/BasicOpenShiftTest.java
```diff
@@ -0,0 +1,102 @@
...
+    private final OpenShiftServer openShiftServer;
+    private final OpenShift openShiftClient;
+
+    public BasicOpenShiftTest() {
+        this.openShiftServer = new OpenShiftServer(false, true);
+        this.openShiftServer.before();
+        // we want to test the XTF OpenShift client, but we need to configure it with the mocked server client
+        // properties
+        OpenShiftClient mockedServerClient = openShiftServer.getOpenshiftClient();
+        this.openShiftClient = OpenShift.get(
```
mnovak1, line 38: I think we'll need some level of abstraction here which will decide whether to use mock or real Openshift cluster. wdyt?

fabiobrz: +1, thanks @mnovak1 - I'll add something.

core/src/test/java/cz/xtf/core/openshift/mocked/smoke/BasicOpenShiftTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Assert;
+import org.junit.jupiter.api.Test;
+
+import cz.xtf.core.openshift.OpenShift;
+import io.fabric8.kubernetes.api.model.ConfigMap;
+import io.fabric8.kubernetes.api.model.ConfigMapBuilder;
```
simkam, line 13: shouldn't it rather test API from XTF? cz.xtf.builder.builders.ConfigMapBuilder? DeploymentConfigBuilder? ...?

fabiobrz: +1 @simkam - here I just sketched the basic idea, I'll take your advice into account for the changes I'll be doing next.

core/src/test/java/cz/xtf/core/openshift/mocked/smoke/BasicOpenShiftTest.java
fabiobrz: Hi @mnovak1 - while thinking about this, I still don't have a strong opinion on the type of criteria we should use to drive this. A XTF property, e.g.: `xtf.test.server.type=(actual|mocked)`? Or a Maven profile? Just brainstorming here.

WDYT?

CC @simkam

simkam: I prefer to use only mock server in unit tests and (if needed) create a new maven module with integration tests running on OpenShift.

fabiobrz: Ah, actually yes, that could be a solution that BTW would allow for driving the two things separately and asynchronously. WDYT @mnovak1 ?

mnovak1: Agreed. I think it's practical solution as running against real instance will be really rare. Also developing tests for both (moc and real openshift) and keep them stable would be hard achieve.

core/src/test/java/cz/xtf/core/openshift/mocked/smoke/BasicOpenShiftTest.java
fabiobrz: Hi @simkam - I pushed changes following your suggestion. Let me know whether this could be eough for a baseline. In case you'd want more tests (i.e. more builders) then I'll keep it as a draft and add them in the next days.

simkam: looks good to me

fabiobrz:
Hi @mnovak1 - can you have a look to see whether this is good to go? @simkam approved already and I think that if we let it in it'll define the pattern and open up to addidtional tests implementation. WDYT?

simkam (approved):

mnovak1 (approved):

PR #485, opened 2022-03-18
change test image url and fail on any error in SkipForConditionTest

simkam:
the main intention is to don't leak internal URL, but some checks were disabled because

- both metadata tests pass correctly only when valid kubeconfig is present locally
- they pass incorrectly without kubeconfig because tests ignore any other Exception message than is configured
- WildFly image doesn't have `architecture` label so it can't pass even with valid kubeconfig


simkam:
@fabiobrz fyi

mnovak1 (approved):

Issue #488, opened 2022-05-30
@SkipFor annotation generates client error if namespace doesn't exists


mcarlett:
If I use `@SkipFor` annotation to match the image metadata, a temporary image stream will be created in the configured namespace, but, if the namespace is not already existing, there will be an error like `Namespace xxxx not found`, exiting the application.

mnovak1:
Ideal would be not need to import image into Openshift to get the metadata.

mnovak1:
However I'm not sure if it's possible.

mcarlett:
I think it is possible using [Skopeo](https://github.com/containers/skopeo) but it is required to be installed (or maybe downloaded/managed as already done with OC binary)

mnovak1:
Major limitation of Skopeo and other tools is that it must run inside Openshift as it might have configured pull secrets for remote docker registry. Basically it must be called from Openshift. I think the current way is the only one which will 100% work. However I've added a comment to PR https://github.com/xtf-cz/xtf/pull/489 to improve it.

simkam:
I'm not sure about this. Isn't it enough to have `cz.xtf.junit5.listeners.ProjectCreator` in `META-INF/services/org.junit.platform.launcher.TestExecutionListener`. There are more extensions that expect `ProjectCreater` loaded first.

mcarlett:
> I'm not sure about this. Isn't it enough to have `cz.xtf.junit5.listeners.ProjectCreator` in `META-INF/services/org.junit.platform.launcher.TestExecutionListener`? There are more extensions that expect `ProjectCreater` loaded first.

yes it would be a nice and clean solution, but if I use an existing test suite that already manages the project lifecycle could be a little bit tricky having 2 different ways to manage the project creation, thanks for the tip!

PR #489, opened 2022-05-30
#488 Create namespace if not exists

mcarlett:

### DIFF
--- a/junit5/src/main/java/cz/xtf/junit5/model/DockerImageMetadata.java
+++ b/junit5/src/main/java/cz/xtf/junit5/model/DockerImageMetadata.java
@@ -16,6 +16,7 @@ import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.waiting.SimpleWaiter;
 import cz.xtf.core.waiting.Waiter;
 import cz.xtf.core.waiting.failfast.FailFastCheck;
+import cz.xtf.junit5.listeners.ProjectCreator;
 import io.fabric8.openshift.api.model.ImageStream;
 import io.fabric8.openshift.api.model.ImageStreamTag;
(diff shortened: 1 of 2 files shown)


junit5/src/main/java/cz/xtf/junit5/model/DockerImageMetadata.java
```diff
@@ -76,6 +77,12 @@ private static DockerImageMetadata getMetadata(OpenShift openShift, Image image)
             return metadataFromTag;
         }
 
+        //create namespace if not exists
+        if (openShift.getProject(openShift.getNamespace()) == null) {
+            openShift.createProjectRequest();
+            OpenShiftWaiters.get(openShift, () -> false).isProjectReady().waitFor();
```
mnovak1, line 83: There should be used the same logic as in:
https://github.com/xtf-cz/xtf/blob/master/junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java#L24-L56

so all secrets are setup. It would make sense to put above logic into separate method in `ProjectCreator` and delegate namespace creation to it.

mnovak1:
@mcarlett Thinking of it. If `DockerImageMetadata` created namespace (and only in this case) then It should delete it as well to avoid unexpected behaviour. Could it be added as well?

mcarlett:
> @mcarlett Thinking of it. If `DockerImageMetadata` created namespace (and only in this case) then It should delete it as well to avoid unexpected behaviour. Could it be added as well?

ok it makes sense, even if removing the project needs to wait for complete removal (correct but slow)

mnovak1 (approved):

mnovak1:
Thanks, merging.

PR #490, opened 2022-05-31
Add new xtf.openshift.namespace.per.testcase property which will exec…

mnovak1:
…ute every test case in its own namespace. Note this property has limitation that consuming test suite must not create static Openshift instances as those are created before name of test case and thus the namespace for the test is known.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,23 @@ xtf.foo.v2.version=1.0.3
 
 Retrieving an instance with this metadata: `Produts.resolve("product");`
 
+#### Run test cases in separate namespaces using `xtf.openshift.namespace.per.testcase` property 
+You can enable running each test case in separate namespace by setting `xtf.openshift.namespace.per.testcase=true`. 
+
+Namespace names follow pattern: "`${xtf.openshift.namespace}`-Test_Case_Name". 
+For example for `xtf.openshift.namespace=testnamespace` and test case `org.test.SmokeTest` it will be `testnamespace-SmokeTest`.
+
+You can limit the length of created namespace by `xtf.openshift.namespace.per.testcase.length.limit` property. By default it's `25` chars. If limit is breached then
+test case name in namespace name is hashed to hold the limit. So namespace name would like `testnamespace-s623jd6332`
+
+**Warning - Limitations**
+
+In case that you're using this feature, consuming test suite must follow those rules to avoid unexpected behaviour when using `cz.xtf.core.openshift.OpenShift` instances:
+
+* Do not create static `cz.xtf.core.openshift.OpenShift` variable like: `public static final OpenShift openshift = Openshifts.master()` on class level. 
+The reason is that during initialization of static instances the test case and corresponsing namespace is not known. To avoid unexpected behaviour `RuntimeException` is thrown.
+* Similarly as above do not create `cz.xtf.core.openshift.OpenShift` variables in static blocks or do not initialize other static variables which creates `cz.xtf.core.openshift.OpenShift` instance.
+
 #### Service Logs Streaming (SLS)
 This feature allows for you to stream the services output while the test is running; this way you can see immediately 
 what is happening inside the cluster.
@@ -215,6 +232,7 @@ following information:
 * _**output**_: the base path where the log stream files - one for each executed test class - will be created. 
   **OPTIONAL**, if not assigned, logs will be streamed to `System.out`. When assigned, XTF will attempt to create the 
   path in case it doesn't exist and default to `System.out` should any error occur. 
+
 
 ###### Usage examples 
 Given what above, enabling SLS for all test classes is possible by executing the following command: 
--- a/core/src/main/java/cz/xtf/core/bm/BuildManager.java
+++ b/core/src/main/java/cz/xtf/core/bm/BuildManager.java
@@ -28,7 +28,7 @@ public class BuildManager {
                 // Otherwise you can see:
                 // $ oc label namespace <name> "label1=foo"
                 // Error from server (Forbidden): namespaces "<name>" is forbidden: User "<user>" cannot patch resource "namespaces" in API group "" in the namespace "<name>"
-                OpenShifts.admin().namespaces().withName(openShift.getNamespace())
+                OpenShifts.admin(openShift.getNamespace()).namespaces().withName(openShift.getNamespace())
                         .edit(new Visitor<NamespaceBuilder>() {
                             @Override
                             public void visit(NamespaceBuilder builder) {
--- a/core/src/main/java/cz/xtf/core/bm/BuildManagers.java
+++ b/core/src/main/java/cz/xtf/core/bm/BuildManagers.java
@@ -1,7 +1,9 @@
 package cz.xtf.core.bm;
 
+import static cz.xtf.core.config.OpenShiftConfig.OPENSHIFT_NAMESPACE;
+
 import cz.xtf.core.config.BuildManagerConfig;
-import cz.xtf.core.config.OpenShiftConfig;
+import cz.xtf.core.config.XTFConfig;
 import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.openshift.OpenShifts;
 import io.fabric8.kubernetes.api.model.Quantity;
@@ -23,7 +25,7 @@ public class BuildManagers {
                     bm = new BuildManager(OpenShifts.master(buildNamespace));
 
                     // If the build namespace is in a separate namespace to "master" namespace, we assume it is a shared namespace
-                    if (!BuildManagerConfig.namespace().equals(OpenShiftConfig.namespace())) {
+                    if (!BuildManagerConfig.namespace().equals(XTFConfig.get(OPENSHIFT_NAMESPACE))) {
                         OpenShift admin = OpenShifts.admin(buildNamespace);
 
                         try {
--- a/core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
+++ b/core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
@@ -2,6 +2,11 @@ package cz.xtf.core.config;
 
 import java.nio.file.Paths;
 
+import org.apache.commons.lang3.StringUtils;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
 public final class OpenShiftConfig {
     public static final String OPENSHIFT_URL = "xtf.openshift.url";
     public static final String OPENSHIFT_TOKEN = "xtf.openshift.token";
@@ -22,6 +27,19 @@ public final class OpenShiftConfig {
     public static final String OPENSHIFT_MASTER_TOKEN = "xtf.openshift.master.token";
     public static final String OPENSHIFT_ROUTE_DOMAIN = "xtf.openshift.route_domain";
     public static final String OPENSHIFT_PULL_SECRET = "xtf.openshift.pullsecret";
+    public static final String OPENSHIFT_NAMESPACE_PER_TESTCASE = "xtf.openshift.namespace.per.testcase";
+    /**
+     * Used only if xtf.openshift.namespace.per.testcase=true - this property can configure its maximum length. This is useful
+     * in case
+     * where namespace is used in first part of URL of route which must have <64 chars length.
+     */
+    public static final String OPENSHIFT_UNIQUE_NAMESPACE_NAME_LENGTH_LIMIT = "xtf.openshift.namespace.per.testcase.length.limit";
+
+    /**
+     * Used only if xtf.openshift.namespace.per.testcase=true - this property configures default maximum length of namespace
+     * name.
+     */
+    private static final String OPENSHIFT_UNIQUE_NAMESPACE_NAME_LENGTH_LIMIT_DEFAULT = "25";
 
     public static String url() {
         return XTFConfig.get(OPENSHIFT_URL);
@@ -48,8 +66,47 @@ public final class OpenShiftConfig {
         return XTFConfig.get(OPENSHIFT_VERSION);
     }
 
+    /**
+     * @return Returns default namespace as defined in xtf.openshift.namespace property or namespace for currently running test
+     *         case when:
+     *         -Dxtf.openshift.namespace.per.testcase=true.
+     *         In case when current thread does not have associated test case (for example when initializing Openshift instance
+     *         in static variable or static block) then exception is thrown.
+     */
     public static String namespace() {
-        return XTFConfig.get(OPENSHIFT_NAMESPACE);
+        if (useNamespacePerTestcase()) {
+            String namespace = TestCaseContext.getRunningTestcase().get(TestCaseContext.getTestCaseForCurrentThread());
+            if (StringUtils.isEmpty(namespace)) {
+                throw new RuntimeException(
+                        "There is no namespace associated with current thread or test case. This can happen in case that OpenShift instance is created in static variable. In this case avoid using static. Or in thread which is not associated with any test case.");
+            }
+            return namespace;
+        } else {
+            return XTFConfig.get(OPENSHIFT_NAMESPACE);
+        }
+    }
+
+    public static boolean useNamespacePerTestcase() {
+        if (StringUtils.isEmpty(XTFConfig.get(OpenShiftConfig.OPENSHIFT_NAMESPACE_PER_TESTCASE))
+                || (StringUtils.isNotEmpty(XTFConfig.get(OpenShiftConfig.OPENSHIFT_NAMESPACE_PER_TESTCASE))
+                        && !Boolean.parseBoolean(XTFConfig.get(OpenShiftConfig.OPENSHIFT_NAMESPACE_PER_TESTCASE)))) {
+            return false;
+        } else {
+            return true;
+        }
+    }
+
+    /**
+     * @return limit on namespace if it's set by -Dxtf.openshift.unique.namespace.name.length.limit property otherwise returns
+     *         -1 which means not set
+     */
+    public static int getNamespaceLengthLimitForUniqueNamespacePerTest() {
+        if (useNamespacePerTestcase()) {
+            return Integer.parseInt(XTFConfig.get(OPENSHIFT_UNIQUE_NAMESPACE_NAME_LENGTH_LIMIT,
+                    OPENSHIFT_UNIQUE_NAMESPACE_NAME_LENGTH_LIMIT_DEFAULT));
+        } else {
+            return -1;
+        }
     }
 
     public static String binaryPath() {
@@ -60,7 +117,7 @@ public final class OpenShiftConfig {
      * Channel configuration for download of OpenShift client from
      * https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/
      * Channels are: stable, latest, fast, candidate
-     * 
+     *
      * @return channel as configured in xtf.openshift.binary.url.channel property, or default 'stable'
      */
     public static String binaryUrlChannelPath() {
--- a/core/src/main/java/cz/xtf/core/config/TestCaseContext.java
+++ b/core/src/main/java/cz/xtf/core/config/TestCaseContext.java
@@ -1,0 +1,47 @@
+package cz.xtf.core.config;
+
+import static cz.xtf.core.config.OpenShiftConfig.OPENSHIFT_NAMESPACE;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class TestCaseContext {
+
+    /**
+     * Used only in case when test cases are executed in different namespaces when:
+     * -Dxtf.openshift.namespace.per.testcase=true
+     *
+     * This allows to track currently running test case for correct namespace mapping. This is used to automatically find
+     * namespace for running test case when
+     * creating {@link cz.xtf.core.openshift.OpenShift} instances.
+     */
+    private static String TEST_CASE_NAME;
+
+    private static final Map<String, String> testcaseToNamespaceMap = new HashMap<String, String>() {
+        {
+            // put default namespace into map to guarantee that it will be deleted with other namespaces
+            put(XTFConfig.get(OPENSHIFT_NAMESPACE), XTFConfig.get(OPENSHIFT_NAMESPACE));
+        }
+    };
+
+    /**
+     * @return test case name associated with current thread or null if not such mapping exists
+     */
+    public static String getTestCaseForCurrentThread() {
+        return TEST_CASE_NAME;
+    }
+
+    public static void setRunningTestCase(String currentlyRunningTestCaseName) {
+        TEST_CASE_NAME = currentlyRunningTestCaseName;
+    }
+
+    /**
+     * Used only in case when test cases are executed in parallel in different namespaces when:
+     * -Dxtf.openshift.namespace.per.testcase=true
+     *
+     * Maps testcase name -> namespace
+     */
+    public static Map<String, String> getRunningTestcase() {
+        return testcaseToNamespaceMap;
+    }
+}
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -27,7 +27,13 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.builder.EqualsBuilder;
 
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
+
+import cz.xtf.core.config.TestCaseContext;
 import cz.xtf.core.config.WaitingConfig;
 import cz.xtf.core.event.EventList;
 import cz.xtf.core.openshift.crd.CustomResourceDefinitionContextProvider;
@@ -40,6 +46,8 @@ import io.fabric8.kubernetes.api.model.Endpoints;
 import io.fabric8.kubernetes.api.model.EnvVar;
 import io.fabric8.kubernetes.api.model.EnvVarBuilder;
 import io.fabric8.kubernetes.api.model.Event;
+import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
+import io.fabric8.kubernetes.api.model.GenericKubernetesResourceList;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesList;
 import io.fabric8.kubernetes.api.model.LocalObjectReferenceBuilder;
@@ -95,6 +103,13 @@ public class OpenShift extends DefaultOpenShiftClient {
     private static volatile String routeSuffix;
 
     public static final String KEEP_LABEL = "xtf.cz/keep";
+
+    /**
+     * Used to cache created Openshift clients for given test case.
+     */
+    public static final Multimap<String, OpenShift> testcaseToOpenshiftClientMap = Multimaps
+            .synchronizedListMultimap(ArrayListMultimap.create());
+
     /**
      * This label is supposed to be used for any resource created by the XTF to easily distinguish which resources have
      * been created by XTF automation.
@@ -120,7 +135,7 @@ public class OpenShift extends DefaultOpenShiftClient {
             openShiftConfig.setNamespace(namespace);
         }
 
-        return new OpenShift(openShiftConfig);
+        return get(openShiftConfig);
     }
 
     public static OpenShift get(Path kubeconfigPath, String namespace) {
@@ -135,7 +150,7 @@ public class OpenShift extends DefaultOpenShiftClient {
                 openShiftConfig.setNamespace(namespace);
             }
 
-            return new OpenShift(openShiftConfig);
+            return get(openShiftConfig);
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
@@ -152,7 +167,7 @@ public class OpenShift extends DefaultOpenShiftClient {
 
         setupTimeouts(openShiftConfig);
 
-        return new OpenShift(openShiftConfig);
+        return get(openShiftConfig);
     }
 
     public static OpenShift get(String masterUrl, String namespace, String token) {
@@ -165,7 +180,27 @@ public class OpenShift extends DefaultOpenShiftClient {
 
         setupTimeouts(openShiftConfig);
 
-        return new OpenShift(openShiftConfig);
+        return get(openShiftConfig);
+    }
+
+    private static OpenShift get(OpenShiftConfig openShiftConfig) {
+        OpenShift openshift;
+        if (testcaseToOpenshiftClientMap.get(TestCaseContext.getTestCaseForCurrentThread()).size() > 0) {
+            // check that such client config already exists in some of the clients
+            Optional<OpenShift> optionalOpenShift = testcaseToOpenshiftClientMap
+                    .get(TestCaseContext.getTestCaseForCurrentThread()).stream()
+                    .filter(oc -> isEqualOpenshiftConfig(openShiftConfig, oc.getConfiguration()))
+                    .findFirst();
+
+            if (optionalOpenShift.isPresent()) {
+                return optionalOpenShift.get();
+            }
+        }
+
+        openshift = new OpenShift(openShiftConfig);
+        testcaseToOpenshiftClientMap.put(TestCaseContext.getTestCaseForCurrentThread(),
+                openshift);
+        return openshift;
     }
 
     private static void setupTimeouts(OpenShiftConfig config) {
@@ -676,6 +711,7 @@ public class OpenShift extends DefaultOpenShiftClient {
                 }
             }
         }
+        log.info("generateHostname returns namespace: " + routeName + "-" + getNamespace() + "." + routeSuffix);
         return routeName + "-" + getNamespace() + "." + routeSuffix;
     }
 
@@ -894,7 +930,8 @@ public class OpenShift extends DefaultOpenShiftClient {
     }
 
     public boolean deleteServiceAccount(ServiceAccount serviceAccount) {
-        return serviceAccounts().delete(serviceAccount);
+        boolean result = serviceAccounts().delete(serviceAccount);
+        return result;
     }
 
     // RoleBindings
@@ -1242,7 +1279,9 @@ public class OpenShift extends DefaultOpenShiftClient {
     public Waiter clean() {
         for (CustomResourceDefinitionContextProvider crdContextProvider : OpenShift.getCRDContextProviders()) {
             try {
-                customResource(crdContextProvider.getContext()).delete(getNamespace());
+                customResources(crdContextProvider.getContext(), GenericKubernetesResource.class,
+                        GenericKubernetesResourceList.class)
+                                .inNamespace(getNamespace()).delete();
                 log.debug("DELETE :: " + crdContextProvider.getContext().getName() + " instances");
             } catch (KubernetesClientException kce) {
                 log.debug(crdContextProvider.getContext().getName() + " might not be installed on the cluster.", kce);
@@ -1267,7 +1306,9 @@ public class OpenShift extends DefaultOpenShiftClient {
         autoscaling().v1().horizontalPodAutoscalers().withLabelNotIn(KEEP_LABEL, "", "true").delete();
         getUserConfigMaps().forEach(this::deleteConfigMap);
         getUserSecrets().forEach(this::deleteSecret);
-        getUserServiceAccounts().forEach(this::deleteServiceAccount);
+        getUserServiceAccounts().forEach((sa) -> {
+            this.deleteServiceAccount(sa);
+        });
         getUserRoleBindings().forEach(this::deleteRoleBinding);
         rbac().roles().withLabelNotIn(KEEP_LABEL, "", "true").withLabelNotIn("olm.owner.kind", "ClusterServiceVersion")
                 .delete();
@@ -1341,5 +1382,16 @@ public class OpenShift extends DefaultOpenShiftClient {
     @Deprecated
     public OpenShiftWaiters waiters() {
         return waiters;
+    }
+
+    private static boolean isEqualOpenshiftConfig(OpenShiftConfig newConfig, OpenShiftConfig existingConfig) {
+        return new EqualsBuilder()
+                .append(newConfig.getOpenShiftUrl(), existingConfig.getOpenShiftUrl())
+                .append(newConfig.getNamespace(), existingConfig.getNamespace())
+                .append(newConfig.getUsername(), existingConfig.getUsername())
+                .append(newConfig.getPassword(), existingConfig.getPassword())
+                .append(newConfig.getRequestConfig().getOauthToken(), existingConfig.getRequestConfig().getOauthToken())
+                .append(newConfig.isTrustCerts(), existingConfig.isTrustCerts())
+                .isEquals();
     }
 }
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
@@ -15,12 +15,17 @@ import cz.xtf.core.waiting.SimpleWaiter;
 import cz.xtf.core.waiting.SupplierWaiter;
 import cz.xtf.core.waiting.Waiter;
 import cz.xtf.core.waiting.failfast.FailFastCheck;
+import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
+import io.fabric8.kubernetes.api.model.GenericKubernetesResourceList;
+import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.openshift.api.model.Build;
 import io.fabric8.openshift.api.model.BuildStatus;
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public class OpenShiftWaiters {
     private OpenShift openShift;
     private FailFastCheck failFast;
@@ -120,16 +125,44 @@ public class OpenShiftWaiters {
     public Waiter isProjectClean() {
         return new SimpleWaiter(() -> {
             int crdInstances = 0;
+            List<GenericKubernetesResource> customResourceDefinitionList = null;
             for (CustomResourceDefinitionContextProvider crdContextProvider : OpenShift.getCRDContextProviders()) {
                 try {
-                    crdInstances += ((List) (openShift.customResource(crdContextProvider.getContext())
-                            .list(openShift.getNamespace()).get("items"))).size();
+                    customResourceDefinitionList = openShift.customResources(crdContextProvider.getContext(),
+                            GenericKubernetesResource.class, GenericKubernetesResourceList.class)
+                            .inNamespace(openShift.getNamespace())
+                            .list().getItems();
+                    crdInstances += customResourceDefinitionList.size();
                 } catch (KubernetesClientException kce) {
                     // CRD might not be installed on the cluster
                 }
             }
-            return crdInstances == 0 & openShift.listRemovableResources().isEmpty();
-        }, TimeUnit.MILLISECONDS, WaitingConfig.timeoutCleanup(), "Cleaning project.")
+
+            boolean isClean = false;
+            List<HasMetadata> listRemovableResources = openShift.listRemovableResources();
+            if (crdInstances == 0 & listRemovableResources.isEmpty()) {
+                isClean = true;
+            } else {
+                StringBuilder strBuilderResourcesToDelete = new StringBuilder(
+                        "Cleaning project - " + openShift.getNamespace()
+                                + " Waiting for following resources to be deleted: \n");
+                if (customResourceDefinitionList != null && !customResourceDefinitionList.isEmpty()) {
+                    customResourceDefinitionList.stream().forEach((r) -> {
+                        strBuilderResourcesToDelete.append(r + "\n");
+                    });
+                }
+                if (!listRemovableResources.isEmpty()) {
+                    listRemovableResources.stream().forEach((r) -> {
+                        strBuilderResourcesToDelete.append(r + "\n");
+                    });
+                }
+                log.debug(strBuilderResourcesToDelete.toString());
+            }
+            return isClean;
+        }, TimeUnit.MILLISECONDS, WaitingConfig.timeoutCleanup(), "Cleaning project - " + openShift.getNamespace())
+                .onTimeout(() -> log.info("Cleaning namespace: " + openShift.getNamespace() + " - timed out."))
+                .onFailure(() -> log.info("Cleaning namespace: " + openShift.getNamespace() + " - failed."))
+                .onSuccess(() -> log.info("Cleaning namespace: " + openShift.getNamespace() + " - finished."))
                 .failFast(failFast);
     }
 
--- a/junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
+++ b/junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
@@ -1,5 +1,7 @@
 package cz.xtf.junit5.listeners;
 
+import static org.junit.platform.engine.TestExecutionResult.Status.FAILED;
+
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -9,9 +11,12 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
+import org.junit.platform.engine.TestExecutionResult;
 import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
 import org.junit.platform.launcher.TestPlan;
 
+import cz.xtf.core.config.OpenShiftConfig;
 import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.openshift.OpenShifts;
 import io.fabric8.kubernetes.api.model.Event;
@@ -21,7 +26,29 @@ import lombok.extern.slf4j.Slf4j;
 public class EventsRecorder implements TestExecutionListener {
     @Override
     public void testPlanExecutionFinished(TestPlan testPlan) {
-        final Path eventsLogPath = Paths.get("log", "events");
+        if (OpenShiftConfig.useNamespacePerTestcase()) {
+            return;
+        }
+        recordEvents(Paths.get("log", "events"));
+    }
+
+    @Override
+    public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult testExecutionResult) {
+        if (OpenShiftConfig.useNamespacePerTestcase() && testIdentifier.isTest()
+                && FAILED.equals(testExecutionResult.getStatus())) {
+            recordEvents(Paths.get("log", "pods", getTestDisplayName(testIdentifier), "events"));
+        }
+    }
+
+    private String getTestDisplayName(TestIdentifier testIdentifier) {
+        String className = testIdentifier.getParentId().get()
+                .replaceAll(".*class:", "")
+                .replaceAll("].*", "");
+        return String.format("%s#%s", className, testIdentifier.getDisplayName());
+    }
+
+    private void recordEvents(Path eventsLogPath) {
+
         final OpenShift openShift = OpenShifts.master();
 
         eventsLogPath.getParent().toFile().mkdirs();
--- a/junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
+++ b/junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
@@ -1,12 +1,28 @@
 package cz.xtf.junit5.listeners;
 
+import static cz.xtf.core.config.OpenShiftConfig.OPENSHIFT_NAMESPACE;
+
+import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
 
+import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.engine.descriptor.MethodBasedTestDescriptor;
+import org.junit.platform.engine.FilterResult;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.launcher.PostDiscoveryFilter;
 import org.junit.platform.launcher.TestExecutionListener;
 import org.junit.platform.launcher.TestPlan;
 
 import cz.xtf.core.config.OpenShiftConfig;
+import cz.xtf.core.config.TestCaseContext;
+import cz.xtf.core.config.XTFConfig;
 import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.openshift.OpenShifts;
 import cz.xtf.core.waiting.SimpleWaiter;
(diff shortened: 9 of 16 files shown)


README.md
```diff
@@ -157,6 +157,23 @@ xtf.foo.v2.version=1.0.3
...
+
+You can limit the length of created namespace by `xtf.openshift.namespace.per.testcase.length.limit` property. By default it's `25` chars. If limit is breached then
+test case name in namespace name is hashed to hold the limit. So namespace name would like `testnamespace-s623jd6332`
+
+**Warning - Limitations**
+
+In case that you're using this feature, consuming test suite must follow those rules to avoid unexpected behaviour when using `cz.xtf.core.openshift.OpenShift` instances:
+
+* Do not create static `cz.xtf.core.openshift.OpenShift` variable like: `public static final OpenShift openshift = Openshifts.master()` on class level. 
+The reason is that during initialization of static instances the test case and corresponsing namespace is not known. To avoid unexpected behaviour `RuntimeException` is thrown.
```
mchoma, line 174: `To avoid unexpected behaviour `RuntimeException` is thrown.` ->

`To avoid unexpected behaviour `RuntimeException` is thrown, when programmer will break this rule`

?

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
```diff
@@ -0,0 +1,47 @@
...
+        {
+            // put default namespace into map to guarantee that it will be deleted with other namespaces
+            put(XTFConfig.get(OPENSHIFT_NAMESPACE), XTFConfig.get(OPENSHIFT_NAMESPACE));
+        }
+    };
+
+    /**
+     * @return test case name associated with current thread or null if not such mapping exists
+     */
+    public static String getTestCaseForCurrentThread() {
```
mchoma, line 30: I must admit I do not understand now how `ForCurrentThread` is achieved. 

But couldn't be ThreadLocal use ? https://www.baeldung.com/java-threadlocal

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
```diff
@@ -0,0 +1,47 @@
...
+     * This allows to track currently running test case for correct namespace mapping. This is used to automatically find
+     * namespace for running test case when
+     * creating {@link cz.xtf.core.openshift.OpenShift} instances.
+     */
+    private static String TEST_CASE_NAME;
+
+    private static final Map<String, String> testcaseToNamespaceMap = new HashMap<String, String>() {
+        {
+            // put default namespace into map to guarantee that it will be deleted with other namespaces
+            put(XTFConfig.get(OPENSHIFT_NAMESPACE), XTFConfig.get(OPENSHIFT_NAMESPACE));
```
mchoma, line 23: I still do not understand whole solution, but I wonder is really 2x `XTFConfig.get(OPENSHIFT_NAMESPACE)` correct here?  Seems to me suspicious.

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
```diff
@@ -0,0 +1,47 @@
...
+     * Used only in case when test cases are executed in different namespaces when:
+     * -Dxtf.openshift.namespace.per.testcase=true
+     *
+     * This allows to track currently running test case for correct namespace mapping. This is used to automatically find
+     * namespace for running test case when
+     * creating {@link cz.xtf.core.openshift.OpenShift} instances.
+     */
+    private static String TEST_CASE_NAME;
+
+    private static final Map<String, String> testcaseToNamespaceMap = new HashMap<String, String>() {
```
mchoma, line 20: We can determine namespace from testcase, right?

So this is just for caching that value.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -95,6 +103,13 @@ public class OpenShift extends DefaultOpenShiftClient {
     private static volatile String routeSuffix;
 
     public static final String KEEP_LABEL = "xtf.cz/keep";
+
+    /**
+     * Used to cache created Openshift clients for given test case.
+     */
+    public static final Multimap<String, OpenShift> testcaseToOpenshiftClientMap = Multimaps
```
mchoma, line 110: Test Case Name is part of OpenShiftConfig, right? Don't we want Map<OpenShiftConfig, OpenShift> cache?

README.md
mnovak1: fixed.

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mnovak1: Sorry, bad name of the method. I've accidently renamed other one. 

I've decided to step away from `ThreadLocal` as it did fully solved using ForkJoinPool in parallel streams.

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mnovak1: I'll double check it's still needed. I've changed the logic in `ProjectCreator` when deleting namespaces so it should not be needed any longer. However the deafult namespace will be still created as it's required to import image for checking metadata when `@SkipFor` annotation is used.

mchoma:
@mnovak1 Maybe I would appreciate some introduction how this is implemented. So I better understand it if this can be simplified even more. Overall idea seems OK to me and this seems on good way. But I feel code could be better organized to be even simpler. I drop some random thoughts, but maybe they are not relevant and I am missing something.

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
```diff
@@ -0,0 +1,47 @@
+package cz.xtf.core.config;
+
+import static cz.xtf.core.config.OpenShiftConfig.OPENSHIFT_NAMESPACE;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class TestCaseContext {
```
mchoma, line 8: My feeling is this new class should have better defined responsibilities. 

As Is it is very dumb. Just holder of 2 fields. Can we expose some API which will better define what is this class for?

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mnovak1: I don't think it would be good to use `OpenShiftConfig` as key here as most instances of `OpenShiftConfig` would differ just by namespace which is retrieved from testcase->namespace map.

mchoma: > most instances of OpenShiftConfig would differ just by namespace
Exactly. 

And I thought I can't reuse OpenShift Client between different testcases(namespaces), because OpenShiftClient usage relies on default namespace which it was created with . We cannot switch namespace in runtime.

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mnovak1: The best would be to have JUnit5 injecting `TestInfo` but it's not able to do so.

mchoma: I see. So I start to understand purpose of this class. It is to hold information about TestCase.

Could that be documented. 

In that case I wonder do we need `Map<String, String> testcaseToNamespaceMap`? Do we need to keep history of all namespaces?

Isn't it enough to keep just one field `private static String testCaseNamespace`?

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mchoma, line 110: `testcaseToOpenshiftClientMap` -> `testCaseToOpenShiftClientMap` ?

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
```diff
@@ -53,17 +106,71 @@ public void visit(NamespaceBuilder builder) {
...
+                TestCaseContext.getTestcaseToNamespaceMap().putIfAbsent(
+                        ((MethodBasedTestDescriptor) testDescriptor).getTestClass().getName(),
+                        getNameSpaceForTestClass(testDescriptor));
+
+            }
+        }
+        return FilterResult.included(testDescriptor.getDisplayName());
+    }
+
+    private String getNameSpaceForTestClass(TestDescriptor testDescriptor) {
```
mchoma, line 152: This is related to other question regarding `TestCaseContext` responsibilities? Have you considered to add this method to `TestCaseContext.java`

mchoma, line 152: `getNameSpaceForTestClass` -> `getNamespaceForTestClass`

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Used only in case when test cases are executed in different namespaces when:
+     * -Dxtf.openshift.namespace.per.testcase=true
+     *
+     * This allows to track currently running test case for correct namespace mapping. This is used to automatically find
+     * namespace for running test case when
+     * creating {@link cz.xtf.core.openshift.OpenShift} instances.
+     */
+    private static String TEST_CASE_NAME;
```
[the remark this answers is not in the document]
mnovak1: Right, fixed

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mnovak1: I've added `testcaseToNamespaceMap` as preparation for parallel run execution. I would like to keep it.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mnovak1: I've changed the logic here. I see using testcase-> openshift is not effective when there is single namespace which is default use as it would cache as many clients as there are test cases. 
So ideal would be to have simple Set<OpenShift> but this on the other hand would not be effective for this feature where each testcase has different namespace so each OpenShift instance creation would iterate over this set.

Having Map<OpenShiftConfig, OpenShift> cache will not work as namespace in `OpenShiftConfig` is calculated dynamically (based on currently running testcase) but having Map<namespace, OpenShift> would work. 

It would be effective for both of the cases above. I'll re-push the logic.

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mnovak1: Ok, I've removed it as it's no longer needed.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: This one is specific for `ProjectCreator` class as it's used just in it when creating names (on namespaces) for test cases.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mchoma: > Having Map<OpenShiftConfig, OpenShift> cache will not work as namespace in OpenShiftConfig is calculated dynamically (based on currently running testcase) but having Map<namespace, OpenShift> would work.

I don't get that argument. Isn't `namespace` also calculated dynamically based on currently running testcase?

Does `Map<namespace, OpenShift>` mean I can have just one OpenShift client for one namespace?

junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
```diff
@@ -21,7 +26,29 @@
...
-        final Path eventsLogPath = Paths.get("log", "events");
+        if (OpenShiftConfig.useNamespacePerTestcase()) {
+            return;
+        }
+        recordEvents(Paths.get("log", "events"));
+    }
+
+    @Override
+    public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult testExecutionResult) {
+        if (OpenShiftConfig.useNamespacePerTestcase() && testIdentifier.isTest()
```
mchoma, line 37: Is it possible to eliminate `if (OpenShiftConfig.useNamespacePerTestcase` conditions?

E.g. can we record events for each test case even for current state?

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mnovak1: `I don't get that argument. Isn't namespace also calculated dynamically based on currently running testcase?`

The reason why `OpenShiftConfig` cannot be used as key in map is that `OpenShiftConfig` instance is calculating namespace dynamically which is derived from currently running testcase name. So once a few test cases are executed it would map to `OpenShift` instances with different namespace than calculated in `OpenShiftConfig`. 

`Does Map<namespace, OpenShift> mean I can have just one OpenShift client for one namespace?`

It's `MultiMap` so one key can have multiple values. This mapping will group all openshift instances (usually the difference will be just whether it's for "master() or admin()") with the same namespace keeping number of newly created and cached `OpenShift` instances at minimum.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: I would like to have a single place from where namespace for running test case can be retrieved. Currently it is `OpenShiftConfig.namespace()`. If `getNameSpaceForTestClass(..)` is added to `TestCaseContext` then it would bypass `OpenShiftConfig.namespace()` logic.

However you're right that `TestCaseContext` responsibilities are not defined. I think it should server just for the purpose of holding test case name but not to be tied with namespace. I think that `Map<String, String> testcaseToNamespaceMap` should not be in `TestCaseContext` and moved to `OpenshiftConfig`. I'll make some changes and push new way for review.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mchoma: >  So once a few test cases are executed it would map to OpenShift instances with different namespace than calculated in OpenShiftConfig.

Please could you leverage more which condition you have in mind? It is still not clear to me. 

I am thinking of 1. paralel run 2. xtf.openshift.namespace.per.testcase=true 3. xtf.openshift.namespace.per.testcase=false, but could not see the problem.

core/src/main/java/cz/xtf/core/bm/BuildManager.java
```diff
@@ -28,7 +28,7 @@ public BuildManager(OpenShift openShift) {
                 // Otherwise you can see:
                 // $ oc label namespace <name> "label1=foo"
                 // Error from server (Forbidden): namespaces "<name>" is forbidden: User "<user>" cannot patch resource "namespaces" in API group "" in the namespace "<name>"
-                OpenShifts.admin().namespaces().withName(openShift.getNamespace())
+                OpenShifts.admin(openShift.getNamespace()).namespaces().withName(openShift.getNamespace())
```
mchoma, line 31: Must caller of `OpenShifts.admin()` provide namespace. Couldn't `OpenShifts.admin()` find out itself correct namespace?

mchoma, line 31: Must caller of `OpenShifts.admin()` provide namespace. Couldn't `OpenShifts.admin()` find out itself correct namespace?

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -1342,4 +1380,15 @@ private Path storeLog(String log, Path dirPath, String fileName) throws IOExcept
     public OpenShiftWaiters waiters() {
         return waiters;
     }
+
+    private static boolean isEqualOpenshiftConfig(OpenShiftConfig newConfig, OpenShiftConfig existingConfig) {
```
mchoma, line 1384: Shouldn't be such `equal()` method comparing two OpenShiftConfig classes moved to OpenShiftConfig class?

mchoma, line 1384: Shouldn't be such `equal()` method comparing two OpenShiftConfig classes moved to OpenShiftConfig class?

core/src/main/java/cz/xtf/core/openshift/OpenShiftWaiters.java
```diff
@@ -120,16 +125,44 @@ public Waiter isProjectReady() {
...
+                            .inNamespace(openShift.getNamespace())
+                            .list().getItems();
+                    crdInstances += customResourceDefinitionList.size();
                 } catch (KubernetesClientException kce) {
                     // CRD might not be installed on the cluster
                 }
             }
-            return crdInstances == 0 & openShift.listRemovableResources().isEmpty();
-        }, TimeUnit.MILLISECONDS, WaitingConfig.timeoutCleanup(), "Cleaning project.")
+
```
mchoma, line 140: I take this is not directly related to  `xtf.openshift.namespace.per.testcase` changes? But just some troubelshooting logging, right?

mchoma, line 140: I take this is not directly related to  `xtf.openshift.namespace.per.testcase` changes? But just some troubelshooting logging, right?

core/src/main/java/cz/xtf/core/bm/BuildManager.java
mchoma: Seems to me it is happening already. https://github.com/xtf-cz/xtf/pull/490/files#diff-9116143614e7b1c4e2f4cf8be51914960b369af36c9ff4e20a46cf68f95cf7cdR33

mchoma: Seems to me it is happening already. https://github.com/xtf-cz/xtf/pull/490/files#diff-9116143614e7b1c4e2f4cf8be51914960b369af36c9ff4e20a46cf68f95cf7cdR33

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mnovak1: > Please could you leverage more which condition you have in mind? It is still not clear to me.

I think `Map<OpenShiftConfig, OpenShift>` would not work. Let's imagine that that Test1 is running in namespace-test1 and openShiftConfig1 -> openshift1 is put into the this map. openShiftConfig1 does not contain variable to store namespace. There is just static `namespace()` method which returns namespace based on currently running test case.  Then Test2 is started and there is openShiftConfig2. Problem I see here is that openShiftConfig2 is equal to openShiftConfig1. So when openshift2 should be created with namespace-test2 then it will actually return cached openshift1 client.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -165,7 +179,25 @@ public static OpenShift get(String masterUrl, String namespace, String token) {
...
+    }
+
+    private static OpenShift get(OpenShiftConfig openShiftConfig) {
+        OpenShift openshift;
+
+        // check whether such a client already exists
+        Optional<OpenShift> optionalOpenShift = namespaceToOpenshiftClientMap
+                .get(openShiftConfig.getNamespace()).stream()
+                .filter(oc -> isEqualOpenshiftConfig(openShiftConfig, oc.getConfiguration()))
+                .findFirst();
```
mchoma, line 192: Can't this introduce some intermittent odd behaviour?

So I have test where I have under one namespace different OpenShift clients - typically `master()` and `admin`. Doesn't `findFirst()` return those clients randomly? 

This brings me back to my objection that `namespaceToOpenshiftClientMap` must use as a key OpenShiftConfig. Then username will be taken into account correctly.

README.md
```diff
@@ -157,6 +157,23 @@ xtf.foo.v2.version=1.0.3
 
 Retrieving an instance with this metadata: `Produts.resolve("product");`
 
+#### Run test cases in separate namespaces using `xtf.openshift.namespace.per.testcase` property 
+You can enable running each test case in separate namespace by setting `xtf.openshift.namespace.per.testcase=true`. 
+
+Namespace names follow pattern: "`${xtf.openshift.namespace}`-Test_Case_Name".
```
mchoma, line 163: `Test_Case_Name` -> `TestCaseName` ?

README.md
```diff
@@ -157,6 +157,23 @@ xtf.foo.v2.version=1.0.3
...
+
+You can limit the length of created namespace by `xtf.openshift.namespace.per.testcase.length.limit` property. By default it's `25` chars. If limit is breached then
+test case name in namespace name is hashed to hold the limit. So namespace name would like `testnamespace-s623jd6332`
+
+**Warning - Limitations**
+
+In case that you're using this feature, consuming test suite must follow those rules to avoid unexpected behaviour when using `cz.xtf.core.openshift.OpenShift` instances:
+
+* Do not create static `cz.xtf.core.openshift.OpenShift` variable like: `public static final OpenShift openshift = Openshifts.master()` on class level. 
+The reason is that during initialization of static instances the test case and corresponsing namespace is not known. To avoid unexpected behaviour `RuntimeException` is thrown, when programmer breaks rule.
```
mchoma, line 174: `when programmer breaks rule` -> `when programmer breaks this rule`

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
```diff
@@ -53,17 +93,71 @@ public void visit(NamespaceBuilder builder) {
...
+        }
+    }
+
+    @Override
+    public FilterResult apply(TestDescriptor testDescriptor) {
+        if (testDescriptor instanceof MethodBasedTestDescriptor) {
+            boolean disabled = Arrays.stream(((MethodBasedTestDescriptor) testDescriptor).getTestClass().getAnnotations())
+                    .filter(annotation -> annotation instanceof Disabled).count() > 0;
+            if (!disabled) {
+                OpenShiftConfig.getTestcaseToNamespaceMap().putIfAbsent(
```
mchoma, line 130: I have strong feeling this can be simplified to new OpenShiftConfig method with TestDescriptor parameter e.g. something like `OpenShiftConfig.matchTestSuiteToNamespace(testDescriptor)` (name can be improved)

That would mean to move `getNamespaceForTestClass` to `OpenShiftConfig`.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mchoma: > openShiftConfig1 does not contain variable to store namespace

Ok, now I see what you mean. 

Alternativelly we can add fields to `OpenShiftConfig`(But it seems it is not how whole package is designed). Or alternativelly use `io.fabric8.openshift.client.OpenShiftConfig` as a key (Not sure if consfusion does not comes there exists same classes in different packages)?

[1] https://github.com/mnovak1/xtf/blob/ff552bba7aef990e3e1f7885d5c1f343d87f44e4/core/src/main/java/cz/xtf/core/openshift/OpenShift.java#L1384

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mnovak1: Can it be set as resolved?

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
mnovak1: Can it be set as resolved?

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mnovak1: Can it be set as resolved?

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: Can it be set as resolved?

junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
mnovak1: Can it be set as resolved?

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: Can it be set as resolved?

mnovak1:
@mchoma I've addressed all the main topics we've discussed:

- Extract logic from ProjectCreator class to different place (NamespaceManager class was created)
- Encapulate the logic not to expose internal maps

core/src/main/java/cz/xtf/core/bm/TestCaseContext.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class TestCaseContext {
+
+    /**
+     *
+     * This allows to track currently running test case for correct namespace mapping. This is used to automatically find
+     * namespace for running test case when
+     * creating {@link cz.xtf.core.openshift.OpenShift} instances.
+     */
+    private static String testCaseName;
```
mchoma, line 15: field and getter/setter does not match

testCaseName vs. runningTestCaseName

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mchoma: Have you considered moving class to `junit5/src/main/java/cz/xtf/junit5/`.

README.md
```diff
@@ -157,6 +157,32 @@ xtf.foo.v2.version=1.0.3
 
 Retrieving an instance with this metadata: `Produts.resolve("product");`
 
+#### Run test cases in separate namespaces using `xtf.openshift.namespace.per.testcase` property 
+
+This feature requires to have XTF JUnit5 `cz.xtf.junit5.listeners.ProjectCreator` extension enabled. This can be done by adding
```
mchoma, line 162: Do we need to duplicate here how to enable ProjectCreator, shouldn't we just link to ProjectCreator section?

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
```diff
@@ -0,0 +1,119 @@
+package cz.xtf.core.bm;
```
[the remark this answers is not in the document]
mchoma: This is still valid question

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
```diff
@@ -0,0 +1,119 @@
...
+        return testcaseToNamespaceMap;
+    }
+
+    /**
+     * Used only in case when test cases are executed in different namespaces when:
+     * -Dxtf.openshift.namespace.per.testcase=true
+     *
+     * @return return namespace for currently running testcase or null if not present
+     */
+    public static String getNamespaceForTestcase(String testcase) {
```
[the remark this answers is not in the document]
mchoma: still valid

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
```diff
@@ -0,0 +1,140 @@
...
+        return getTestcaseToNamespaceMap().get(testcase);
+    }
+
+    public static boolean createIfDoesNotExistsProject(String namespace) {
```
mchoma, line 55: Why can't be this method left in `ProjectCreator`?

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
```diff
@@ -0,0 +1,140 @@
...
+            isCreated = true;
+        }
+        if (OpenShiftConfig.pullSecret() != null) {
+            openShift.setupPullSecret(OpenShiftConfig.pullSecret());
+        }
+        log.info("Created namespace: " + openShift.getNamespace());
+        return isCreated;
+    }
+
+    public static boolean deleteProject(String namespace, boolean waitForDeletion) {
```
mchoma, line 96: Why can't be this method left in `ProjectCreator`?

core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
```diff
@@ -48,8 +68,44 @@ public static String version() {
...
+    /**
+     * @return Returns default namespace as defined in xtf.openshift.namespace property or namespace for currently running test
+     *         case when:
+     *         -Dxtf.openshift.namespace.per.testcase=true.
+     *         In case when current thread does not have associated test case (for example when initializing Openshift instance
+     *         in static variable or static block) then exception is thrown.
+     */
     public static String namespace() {
-        return XTFConfig.get(OPENSHIFT_NAMESPACE);
+        if (useNamespacePerTestcase()) {
```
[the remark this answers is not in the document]
mchoma: still valid

junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
```diff
@@ -9,19 +11,44 @@
...
 
+import cz.xtf.core.config.OpenShiftConfig;
 import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.openshift.OpenShifts;
 import io.fabric8.kubernetes.api.model.Event;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class EventsRecorder implements TestExecutionListener {
+    // TODO consider to remove this method as having namespace as test case should be considered as default and single namesapce as special case
```
mchoma, line 27: This TODO in scope of this MR?

mchoma (changes requested):

junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
mchoma: Yes, as it was addressed.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mchoma: Yes, my understanding is this was addressed.

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mnovak1: Problem is that it would create dependency from core -> junit5 module. I would like to avoid it and better have the opposite. Let other modules depend on core.

core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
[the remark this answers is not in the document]
mnovak1: Thinking of it, It would introduce another entry point where namespace can be retrieved. I wanted to have just one place where this information could be retrieved. Now it's `OpenShiftConfig.namespace()`. I don't think there is a room to change it as moving it to `NamespaceManager.getNamespace()` would require refactoring/changes in consuming test suite.

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
```diff
@@ -0,0 +1,119 @@
...
+     * Used only in case when test cases are executed in different namespaces when:
+     * -Dxtf.openshift.namespace.per.testcase=true
+     *
+     * @return return namespace for currently running testcase or null if not present
+     */
+    public static String getNamespaceForTestcase(String testcase) {
+        return getTestcaseToNamespaceMap().get(testcase);
+    }
+
+    public static void createIfDoesNotExistsProject(String namespace) {
```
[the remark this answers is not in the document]
mnovak1: `NamespaceManager` puts logic for creating namespace to one place so it can be called not just by `ProjectCreator`. For example it's used in `DockerImageMetadata`: https://github.com/xtf-cz/xtf/pull/490/files#diff-d4e5929b72c1d68c1e28f2ca9827840373ad1cb557aed670eeeefedd0400d7afR84

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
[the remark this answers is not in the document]
mnovak1: Right, I've created new packages for them.

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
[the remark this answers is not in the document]
mnovak1: This is related to discussion above to have single place of truth where to get namespace. If added then all tests should get namespace from here and it would mean lots of changes in consuming test suites.

core/src/main/java/cz/xtf/core/bm/TestCaseContext.java
mnovak1: Right, fixed.

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
mnovak1: Moving this logic to `NamespaceManager` provides single place for creating namespace.

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
mnovak1: Same reason as above.

junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
mnovak1: This would break current behaviour. I think we can do that when moving to XTF 2.0 one day

core/src/main/java/cz/xtf/core/config/TestCaseContext.java
mchoma: Makes totally sense. Resolving.

junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
mchoma: We are still not even 1.0, right? :)

We do not have semantic versioning guarantees on not changing behaviour, do we? It is our own constraint limitation? Do we gather those breaking changes for next major release?

README.md
mchoma: Will you create it?

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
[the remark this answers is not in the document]
mchoma: Ok. To be honest I do not understand why DockerImageMetadata should create project, but I don't want to go to that rabbit hole now :) It is your decision and I will accept it. 

Resolving this then.

README.md
mnovak1: yes :-)

core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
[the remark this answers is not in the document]
mnovak1: I guess we'll need to make this change sooner or later. However we need to behave correctly like if we're at 1.0 version as XTF is consumed by number of teams and they would not be happy if anything breaks. 

What about to deprecate `OpenShiftConfig.namespace()` and use `NamespaceManager.getNamespace()` instead. And make moving consuming TSs to `NamespaceManager.getNamespace()` are prerequisite for this feature.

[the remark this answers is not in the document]
mchoma: > What about to deprecate `OpenShiftConfig.namespace()` and use `NamespaceManager.getNamespace()` instead. And make moving consuming TSs to `NamespaceManager.getNamespace()` are prerequisite for this feature.

I agree.

README.md
```diff
@@ -157,16 +157,21 @@ xtf.foo.v2.version=1.0.3
...
 This feature requires to have XTF JUnit5 `cz.xtf.junit5.listeners.ProjectCreator` extension enabled. This can be done by adding
-`cz.xtf.junit5.listeners.ProjectCreator` into files:
+`cz.xtf.junit5.listeners.ProjectCreator` line into files:
 ```
 src/test/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
 src/test/resources/META-INF/services/org.junit.platform.launcher.PostDiscoveryFilter
 src/test/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
 ```
 
+#### Run test cases in separate namespaces using `xtf.openshift.namespace.per.testcase` property
```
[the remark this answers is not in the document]
mnovak1: Right, I've added this information.

core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
```diff
@@ -69,23 +68,14 @@ public static String version() {
...
-     *         case when:
-     *         -Dxtf.openshift.namespace.per.testcase=true.
-     *         In case when current thread does not have associated test case (for example when initializing Openshift instance
-     *         in static variable or static block) then exception is thrown.
+     * @deprecated
+     * @see NamespaceManager#getNamespace()
+     *
+     * @return Returns default namespace as defined in xtf.openshift.namespace property
      */
+    @Deprecated
```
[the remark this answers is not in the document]
mnovak1: You're right. However there will be problem with current mindset that `OpenShiftConfig` is the source of thruth which is actually the reason why I placed in `OpenShiftConfig` at the first place. 

If you're ok then I'll remove the deprecation and just mention in the javadoc that `NamespaceManager.getNamespace()` is most likely what they want to use.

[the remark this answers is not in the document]
mchoma: I agree just good javadoc description of methods `NamespaceManager.getNamespace()` and `OpenShiftConfig.getNamespace` is solving this (+ README note, which is different thread about)

mchoma:
@mnovak1 after we will resolve all threads (hopefully in near future) we will need some final CI run to check everything is still OK, as this MR was refactored significantly compared to origin version.

core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
```diff
@@ -2,6 +2,12 @@
 
 import java.nio.file.Paths;
 
+import org.apache.commons.lang3.StringUtils;
+
+import cz.xtf.core.namespace.NamespaceManager;
```
mchoma, line 7: Unused import?

If yes, could you double check all classes please?

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
```diff
@@ -0,0 +1,194 @@
...
+                        + StringUtils.truncate(DigestUtils.sha256Hex(testDescriptor.getParent().get().getDisplayName()
+                                .toLowerCase()),
+                                OpenShiftConfig.getNamespaceLengthLimitForUniqueNamespacePerTest()
+                                        - XTFConfig.get(OPENSHIFT_NAMESPACE).length());
+            } else {
+                return XTFConfig.get(OPENSHIFT_NAMESPACE) + "-"
+                        + testDescriptor.getParent().get().getDisplayName().toLowerCase(); // namespace must not have upper case letters
+            }
+        } else {
+            return XTFConfig.get(OPENSHIFT_NAMESPACE);
```
mchoma, line 166: `OpenShiftConfig.getNamespace()` can be used?

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
```diff
@@ -0,0 +1,194 @@
...
+    public static String getNamespace() {
+        if (OpenShiftConfig.useNamespacePerTestcase()) {
+            String namespace = NamespaceManager.getNamespaceForTestcase(TestCaseContext.getRunningTestCaseName());
+            if (StringUtils.isEmpty(namespace)) {
+                throw new RuntimeException(
+                        "There is no namespace associated with current thread or test case. This can happen in case that OpenShift instance is created in static variable. In this case avoid using static. Or in thread which is not associated with any test case.");
+            }
+            return namespace;
+        } else {
+            return XTFConfig.get(OPENSHIFT_NAMESPACE);
```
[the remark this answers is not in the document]
mchoma: Maybe check all occurences of `XTFConfig.get(OPENSHIFT_NAMESPACE)`

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * @return Returns default namespace as defined in xtf.openshift.namespace property or namespace for currently running test
+     *         case when:
+     *         -Dxtf.openshift.namespace.per.testcase=true.
+     *         In case when current thread does not have associated test case (for example when initializing Openshift instance
+     *         in static variable or static block) then {@link java.lang.RuntimeException} exception is thrown.
+     */
+    public static String getNamespace() {
+        if (OpenShiftConfig.useNamespacePerTestcase()) {
```
mchoma, line 183: Do we need this if here when same if/else construction is in `getNamespaceForTestcase` ?

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
```diff
@@ -1,92 +1,76 @@
...
-     */
-    public static boolean deleteProject() {
-        boolean deleted = openShift.deleteProject();
-        // For multi-module maven projects, other modules may attempt to crate project requests immediately after this modules deleteProject
-        if (deleted) {
-            BooleanSupplier bs = () -> openShift.getProject() == null;
-            new SimpleWaiter(bs, TimeUnit.MINUTES, 2, "Waiting for old project deletion").waitFor();
+    @Override
+    public void afterAll(ExtensionContext context) {
+        if (OpenShiftConfig.useNamespacePerTestcase() && JUnitConfig.cleanOpenShift()) {
```
mchoma, line 53: Couldn't be `OpenShiftConfig.useNamespacePerTestcase()` logic hidden in `NamespaceManager.deleteProject()` ? 

Seems to me yes. Just delete `NamespaceManager.getNamespace()` and `NamespaceManager.getNamespace()` will return correct namespace to delete.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
```diff
@@ -1,92 +1,76 @@
...
+        NamespaceManager.createIfDoesNotExistsProject(XTFConfig.get(OPENSHIFT_NAMESPACE));
     }
 
     @Override
-    public void testPlanExecutionFinished(TestPlan testPlan) {
-        if (JUnitConfig.cleanOpenShift()) {
-            deleteProject();
-        }
+    public void beforeAll(ExtensionContext context) {
+        // todo consider to move setting testcase context in separate extension
```
mchoma, line 37: This is good point. Especially when you say there si at least one team which does not use ProjectCreator.

I think such new Extension could be registered by default.

README.md
```diff
@@ -157,6 +159,39 @@ xtf.foo.v2.version=1.0.3
...
+
+Namespace names follow pattern: "`${xtf.openshift.namespace}`-TestCaseName". 
+For example for `xtf.openshift.namespace=testnamespace` and test case `org.test.SmokeTest` it will be `testnamespace-SmokeTest`.
+
+You can limit the length of created namespace by `xtf.openshift.namespace.per.testcase.length.limit` property. By default it's `25` chars. If limit is breached then
+test case name in namespace name is hashed to hold the limit. So namespace name would like `testnamespace-s623jd6332`
+
+**Warning - Limitations**
+
+For this feature to work, **you must not use `OpenShiftConfig.getNamespace()` and use `NamespaceManager.getNamespace()` instead.**
```
mchoma, line 187:
```
When enabling this feature in your project, **you may need to replace `OpenShiftConfig.getNamespace()` with`NamespaceManager.getNamespace()`. Check method's javadoc to understand difference.** 
```
?

core/src/main/java/cz/xtf/core/bm/BuildManagers.java
```diff
@@ -23,7 +25,7 @@ public static BuildManager get() {
                     bm = new BuildManager(OpenShifts.master(buildNamespace));
 
                     // If the build namespace is in a separate namespace to "master" namespace, we assume it is a shared namespace
-                    if (!BuildManagerConfig.namespace().equals(OpenShiftConfig.namespace())) {
+                    if (!BuildManagerConfig.namespace().equals(XTFConfig.get(OPENSHIFT_NAMESPACE))) {
```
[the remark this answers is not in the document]
mnovak1: Right, I'll change it.

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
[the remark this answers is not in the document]
mnovak1: Fixed.

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
mnovak1: fixed

core/src/main/java/cz/xtf/core/context/TestCaseContext.java
```diff
@@ -0,0 +1,28 @@
+package cz.xtf.core.context;
+
+/**
+ * Used only in case when test cases are executed in different namespaces when:
```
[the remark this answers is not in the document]
mnovak1: Old javadoc, removed. Test case context is set no matter what is set in `xtf.openshift.namespace.per.testcase` property.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: I was looking at and it does not seems to be possible to register extension by default in consuming test suite.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: Right, fixed.

README.md
mnovak1: fixed

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
mchoma: I mean this https://github.com/xtf-cz/xtf/pull/490/files#diff-5b54c74eced1a279848d833e9dec30d13ae93e5f31a60c45a06298d85f4de130R143

mchoma: Oh, but it is different method :) 

`getNamespaceForTestClass` vs. `getNamespaceForTestcase`

mchoma: But you got my point of eliminating `if (OpenShiftConfig.useNamespacePerTestcase()) {`, right? :)

Can't  `NamespaceManager.getNamespaceForTestcase(TestCaseContext.getRunningTestCaseName());` work for `else` codepath as well?

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
```diff
@@ -0,0 +1,191 @@
...
+        return testcaseToNamespaceMap;
+    }
+
+    /**
+     * Used only in case when test cases are executed in different namespaces when:
+     * -Dxtf.openshift.namespace.per.testcase=true
+     *
+     * @return return namespace for currently running testcase or null if not present
+     */
+    public static String getNamespaceForTestcase(String testcase) {
```
mchoma, line 49: I see you are using in this MR both `Testcase` and `TestCase` inconsistly. 

My personal attitude is `TestCase` is correct one.

junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
```diff
@@ -9,19 +11,44 @@
...
 import io.fabric8.kubernetes.api.model.Event;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class EventsRecorder implements TestExecutionListener {
+    // TODO consider to remove this method as having namespace per test case should be considered default and single namespace as special case
     @Override
     public void testPlanExecutionFinished(TestPlan testPlan) {
-        final Path eventsLogPath = Paths.get("log", "events");
+        if (OpenShiftConfig.useNamespacePerTestcase()) {
```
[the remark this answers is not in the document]
mnovak1: I suspect the issue will be that namespace is removed in `@AfterAll` phase and if then tries to record events then it throws exception becuase namespace does not exist. 

But I can add check into `recordEvents()` method that namespace exists and skip recording events there if not. Then this one could be removed.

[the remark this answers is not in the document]
mnovak1: I've added the changes.

core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java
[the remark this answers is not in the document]
mchoma: This was addressed. Resolving.

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * @return Returns default namespace as defined in xtf.openshift.namespace property or namespace for currently running test
+     *         case when:
+     *         -Dxtf.openshift.namespace.per.testcase=true.
+     *         In case when current thread does not have associated test case (for example when initializing Openshift instance
+     *         in static variable or static block) then {@link java.lang.RuntimeException} exception is thrown.
+     */
+    public static String getNamespace() {
+        if (OpenShiftConfig.useNamespacePerTestCase()) {
```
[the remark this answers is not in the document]
mchoma: I agree if we want to guard user do not use Namespace For Test Case in static code with `RuntimeException` we cant avoid this `if`.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: Well, looks like it can be automatically added if consuming test suite has:
`junit.jupiter.extensions.autodetection.enabled=true` (be default it's `false`)

In this case consuming test suite does not need to add it in `META-INF/services/org.junit.jupiter.api.extension.Extension` as it can be registered as extension directly in `cz.xtf.junit5` maven depdency.

However even though it seems to be cleaner design, setting `TestCaseContext` in separate extension must happen before `ProjectCreator` is called. I've tried the above appoach and `ProjectCreator` is called first, before `TestCaseContext` is set which leads to `RuntimeException`. 

Only solution seems to have both. Setting it in extension and here.

mnovak1: Do you have any opinion on this?

As `TestCaseContext` is public I tend to add this extension into `core` module as it's used there and publicly accesible. So use both - setting in `ProjectCreator` and in extension. wdyt?

mchoma: Introduce own Extension sounds to me correct.

Need to keep it here in ProjectCreator seems to me as workaround. Isn't there way to ensure filling `TestCaseContext` is called as very first code when running TestCase?

mchoma: Or simply document in Readme that to properly configure this feature `TestCaseContextExtension` must be registered first?

mnovak1: > Need to keep it here in ProjectCreator seems to me as workaround. Isn't there way to ensure filling TestCaseContext is called as very first code when running TestCase?

Only way seems to be over extension. I didn't find any other way. 

> Or simply document in Readme that to properly configure this feature TestCaseContextExtension must be registered first?

I'll document it but I think having it here will avoid user issues wondering why it's not working. 

I've added it to both places.

mnovak1:
@mchoma I've addressed all the last comments and added CI runs. Could you review it. please?

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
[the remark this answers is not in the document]
mchoma: As `public static String getNamespace()` was introduced, `public static String getNamespaceForTestcase(String testcase) {`  could become private?

junit5/src/main/java/cz/xtf/junit5/listeners/EventsRecorder.java
mchoma: This could be resolved?

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mchoma: I do not see it documented. Am I missing something?

core/src/main/java/cz/xtf/core/bm/NamespaceManager.java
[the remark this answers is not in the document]
mchoma: Resolving.

core/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
```diff
@@ -0,0 +1,2 @@
+cz.xtf.core.context.TestCaseContextExtension
```
mchoma, line 1: So this is always registered?

According to https://github.com/xtf-cz/xtf/pull/490/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R164 it should be registered only when condition is success.

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
```diff
@@ -0,0 +1,192 @@
...
+import cz.xtf.core.waiting.SimpleWaiter;
+import io.fabric8.kubernetes.api.builder.Visitor;
+import io.fabric8.kubernetes.api.model.NamespaceBuilder;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class NamespaceManager {
+    /**
+     * By default (xtf.openshift.namespace.per.testcase=false) all entries in map point to value returned by
```
mchoma, line 26: Do we need to elaborate implementation details here? Cant we just say it is mapping of namespace to testcase.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: Unfortunately `ProjectCreator` is registered and invoked first so it does not solve the problem.

core/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
mnovak1: As mentioned before automatic registration of extensions is not well possible with JUnit 5. `junit.jupiter.extensions.autodetection.enabled=true` must be set for `TestCaseContextExtension` to be automatically registered.

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
mnovak1: It's high level describtion of the logic. It actually does not go into detail from my pov :-)

core/src/main/resources/META-INF/services/org.junit.jupiter.api.extension.Extension
mchoma: Oh, sorry. I misread the README. Documentation is OK. Resolving.

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
mchoma: Ok. Not insisting on that. 

My only concern was to eliminate mention of `xtf.openshift.namespace.per.testcase` to have it encapsulated on few places. But it is still NamespaceManager where it belongs, so resolving.

README.md
```diff
@@ -157,6 +159,50 @@ xtf.foo.v2.version=1.0.3
...
+
+Namespace names follow pattern: "`${xtf.openshift.namespace}`-TestCaseName". 
+For example for `xtf.openshift.namespace=testnamespace` and test case `org.test.SmokeTest` it will be `testnamespace-SmokeTest`.
+
+You can limit the length of created namespace by `xtf.openshift.namespace.per.testcase.length.limit` property. By default it's `25` chars. If limit is breached then
+test case name in namespace name is hashed to hold the limit. So namespace name would like `testnamespace-s623jd6332`
+
+**Warning - Limitations**
+
+When enabling this feature in your project, **you may need to replace `OpenShiftConfig.getNamespace()` with`NamespaceManager.getNamespace()`. Check method's javadoc to understand difference.**
```
mchoma, line 198: Nice to have. Is it possible to provide links to: 
- `OpenShiftConfig.getNamespace()` 
- `NamespaceManager.getNamespace()`

So reader can check javadoc as reads.

mchoma:
Ok I run through changes and I do not have any signifacnt objection. But truth is I am looking for long time on MR. Maybe it would need some fresh view :).

Just to double check. For someone who is not interested in using `xtf.openshift.namespace.per.testcase=true` it has no migration consequences?

README.md
mnovak1: Right, fixed.

mnovak1:
> Just to double check. For someone who is not interested in using xtf.openshift.namespace.per.testcase=true it has no migration consequences?

Besides use of the static OpenShift instance and use of NamespaceManager there is no migration issue.

README.md
mchoma: Fixed. Resolving.

simkam (approved):
TBH I think that mapping should be done at a different level. TestCase to namespaced OpenShiftClient rather than TestCase to OpenShiftClient to namespace. 

But there might be reasons to go this way. 

Quite a lot of static fields might also prove to be problematic in the future. 

Anyway, approving.

PR #504, opened 2022-08-24
Fix OpenShiftRecorderService in order to create a unique directory for parametrized tests

### LINKED ISSUE #501: If parameterized tests fail then OpenShiftRecorderService throws java.nio.file.FileAlreadyExistsException when overriding files from build namespace
In case a test fails `OpenShiftRecorderService` saves all files related to test (including from build namespace) to "resources" directory. Problem is with parametrized tests (annotated by `@ParameterizedTest`) as all of them are sharing the same directory and thus trying to override the same files. 

For example following directories are created:
```
	com.redhat.xpaas.eap.template.StaticTest.[1] eap74-https-s2i
	com.redhat.xpaas.eap.template.StaticTest.[2] eap74-basic-s2i
	com.redhat.xpaas.eap.template.StaticTest.[3] eap74-amq-persistent-s2i
	com.redhat.xpaas.eap.template.StaticTest.[4] eap74-amq-s2i
	com.redhat.xpaas.eap.template.StaticTest.[5] eap74-sso-s2i
```
It seems that test name is missing in the name of directory (test case name and parametrization is the name of directory).

fabiobrz:
Adds the test method name to the generated directory name for parametrized tests, so that files can be downloaded without overriding existing ones.

FYI @mnovak1 (CC @simkam) - I am keeping this as a draft while I collect some runs to check the changes. Feel free to let me know what you think.
 
Fix #501 

Local execution with changes shows that now the test name is now used to compute the output directory name, so it will be unique for parametrized tests too:

```
ls -l junit5/
total 32
drwxrwxr-x. 2 fburzigo fburzigo 1024 Aug 29 13:41 cz.xtf.junit5.extensions.OpenShiftRecorderTest.failingParameterTest[2] 1
drwxrwxr-x. 2 fburzigo fburzigo 1024 Aug 29 13:41 cz.xtf.junit5.extensions.OpenShiftRecorderTest.failingParameterTest[3] 2
drwxrwxr-x. 2 fburzigo fburzigo 1024 Aug 29 13:42 cz.xtf.junit5.extensions.OpenShiftRecorderTest.yetAnotherFailingParameterTest[2] 1
drwxrwxr-x. 2 fburzigo fburzigo 1024 Aug 29 13:42 cz.xtf.junit5.extensions.OpenShiftRecorderTest.yetAnotherFailingParameterTest[3] 2
...
```

The alternative conditional branch in `dirNameForTest()` has been fixed too, in order to add a random unique id to the directory name even in the case `context.getTestMethod()` is not defined.  

---

### DIFF
--- a/junit5/src/main/java/cz/xtf/junit5/extensions/OpenShiftRecorderService.java
+++ b/junit5/src/main/java/cz/xtf/junit5/extensions/OpenShiftRecorderService.java
@@ -3,6 +3,7 @@ package cz.xtf.junit5.extensions;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 
@@ -532,9 +533,10 @@ public class OpenShiftRecorderService {
     private String dirNameForTest(ExtensionContext context) {
         // if is test
         if (context.getTestMethod().isPresent()) {
-            return context.getTestClass().get().getName() + "." + context.getDisplayName();
+            return context.getTestClass().get().getName() + "." + context.getTestMethod().get().getName()
+                    + context.getDisplayName();
         } else {
-            return context.getTestClass().get().getName();
+            return context.getTestClass().get().getName() + "-" + ThreadLocalRandom.current().nextInt(Short.MAX_VALUE);
         }
     }
 }


junit5/src/main/java/cz/xtf/junit5/extensions/OpenShiftRecorderService.java
```diff
@@ -534,7 +535,7 @@ private String dirNameForTest(ExtensionContext context) {
         if (context.getTestMethod().isPresent()) {
             return context.getTestClass().get().getName() + "." + context.getDisplayName();
         } else {
-            return context.getTestClass().get().getName();
+            return context.getTestClass().get().getName() + "-" + UUID.randomUUID();
```
mnovak1, line 538: The whole uuid seems long, can it be shortened to just few letters?

fabiobrz: I can't guess what's the use for generating a UUID and then shortening it, i.e. losing its format, one of the ingredients that make the recipe unique. Hence I went for `ThreadLocalRandom.current().nextInt()`.

@mnovak1 let me know whether this works for you.

mnovak1: Nice, I wasn't even aware of `ThreadLocalRandom` class. Reading JavaDoc it seems to be good for the job. Only problem might be with negative values which does not look good. However `ThreadLocalRandom.nextInt(int bound)` seems to return only positive numbers based on JavaDoc. Can it be used?

fabiobrz: Right, good catch! Fixed.

mnovak1 (approved):

mnovak1:
Thanks, merged.

PR #505, opened 2022-08-25
Skip builds also for disabled test classes

jbliznak:
Before this PR, when class is annotated with `@SkipFor` or `@SinceVersion` and these annotations result in test class being skipped, we also skip pre-building of configured builds for that class (introduced in https://github.com/xtf-cz/xtf/pull/277).

This PR extends the check to also skip the build when test class is annotated with JUnit `@Disabled` annotation which further optimizes the execution.

Also some tests were added to verify the previous and enhanced behaviour.

### DIFF
--- a/junit5/pom.xml
+++ b/junit5/pom.xml
@@ -44,6 +44,12 @@
         </dependency>
 
         <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>uk.org.webcompere</groupId>
             <artifactId>system-stubs-jupiter</artifactId>
             <scope>test</scope>
--- a/junit5/src/test/java/cz/xtf/junit5/listeners/ManagedBuildPrebuilderTest.java
+++ b/junit5/src/test/java/cz/xtf/junit5/listeners/ManagedBuildPrebuilderTest.java
@@ -1,0 +1,105 @@
+package cz.xtf.junit5.listeners;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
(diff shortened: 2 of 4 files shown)


junit5/src/test/java/cz/xtf/junit5/listeners/ManagedBuildPrebuilderTest.java
```diff
@@ -0,0 +1,116 @@
...
+@ExtendWith(SystemStubsExtension.class)
+class ManagedBuildPrebuilderTest {
+
+    @SystemStub
+    private SystemProperties systemProperties;
+
+    @BeforeEach
+    public void before() {
+        systemProperties.set("xtf.eap.subid", "test");
+        systemProperties.set("xtf.eap.test.image", "quay.io/wildfly/wildfly-s2i-jdk11:26.0");
```
mnovak1, line 33: Optional: Maybe adding a comment here that if wildfly version is updated then tests need to be updated as well.

simkam (approved):

junit5/pom.xml
```diff
@@ -43,6 +43,12 @@
             <artifactId>assertj-core</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
jbliznak: It is just a test dependency (so no change to public API dependency tree) to avoid SFL4J warning that no impl is available.
We are already using it in other module https://github.com/xtf-cz/xtf/blob/master/core/pom.xml#L76-L80

junit5/src/test/java/cz/xtf/junit5/listeners/ManagedBuildPrebuilderTest.java
jbliznak: comment added

mnovak1 (approved):

PR #510, opened 2022-09-13
Fix namespace deletion

mnovak1:
Namespace deletion between test cases causes intermittent issues. If namespace is created before it's actually deleted then creation of namespace looks like as namespace already exists and it's deleted a few seconds later causing unexpected issues. 

Unfortunately it looks like that kubernetes-client API does not provide reliable way to detect deleted namespace. Thus moving back to safe point and if `xtf.openshift.namespace.per.testcase=false` which is by default do not delete namespace. Only if `xtf.openshift.namespace.per.testcase=true` then delete namespace with nowait.

Also make deletion of namespace using `NamespaceManager.deleteProject(...)` more robust.

### DIFF
--- a/core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
+++ b/core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
@@ -16,6 +16,7 @@ import cz.xtf.core.openshift.OpenShift;
 import cz.xtf.core.openshift.OpenShifts;
 import cz.xtf.core.waiting.SimpleWaiter;
 import io.fabric8.kubernetes.api.builder.Visitor;
+import io.fabric8.kubernetes.api.model.Namespace;
 import io.fabric8.kubernetes.api.model.NamespaceBuilder;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import lombok.extern.slf4j.Slf4j;
@@ -58,12 +59,14 @@ public class NamespaceManager {
     /**
      * Creates namespace if it does not exist
      *
-     * @return true if successful, false otherwise
+     * @return true if newly created, false otherwise (failed or namespace already present)
      */
     public static boolean createIfDoesNotExistsProject(String namespace) {
         OpenShift openShift = OpenShifts.master(namespace);
 
-        boolean isCreated = false;
+        // in case namespace is terminating (means is being deleted) then wait
+        checkAndWaitIfNamespaceIsTerminating(namespace);
+
         if (openShift.getProject() == null) {
             log.info("Creating namespace: " + openShift.getNamespace());
             openShift.createProjectRequest();
@@ -92,13 +95,21 @@ public class NamespaceManager {
                         + openShift.getNamespace() + "'. Possible cause are insufficient permissions.");
                 log.debug(e.getMessage());
             }
-            isCreated = true;
+
+            if (OpenShiftConfig.pullSecret() != null) {
+                openShift.setupPullSecret(OpenShiftConfig.pullSecret());
+            }
+            log.info("Created namespace: " + openShift.getNamespace());
+            return true;
         }
-        if (OpenShiftConfig.pullSecret() != null) {
-            openShift.setupPullSecret(OpenShiftConfig.pullSecret());
+        return false;
+    }
+
+    private static void checkAndWaitIfNamespaceIsTerminating(String namespace) {
+        Namespace n = OpenShifts.admin(namespace).namespaces().withName(namespace).get();
+        if (n != null && n.getStatus().getPhase().equals("Terminating")) {
+            waitForNamespaceToBeDeleted(namespace);
         }
-        log.info("Created namespace: " + openShift.getNamespace());
-        return isCreated;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
```diff
@@ -47,7 +47,7 @@ private void setTestExecutionContext(ExtensionContext context) {
 
     @Override
     public void afterAll(ExtensionContext context) {
-        if (JUnitConfig.cleanOpenShift()) {
+        if (JUnitConfig.cleanOpenShift() && OpenShiftConfig.useNamespacePerTestCase()) {
```
mchoma, line 50: You know I was against introducing those `if OpenShiftConfig.useNamespacePerTestCase()`.

I do not understand why `if xtf.openshift.namespace.per.testcase=false` project is not deleted. Isn't it changing current behaviour?

mchoma:
> Namespace deletion between test cases causes intermittent issues. If namespace is created before it's actually deleted then creation of namespace looks like as namespace already exists and it's deleted a few seconds later causing unexpected issues.

I do not understand why it is happening. Shouldn't be namespace for test case unique?

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
```diff
@@ -58,12 +59,14 @@ public static boolean createIfDoesNotExistsProject() {
...
      * Creates namespace if it does not exist
      *
-     * @return true if successful, false otherwise
+     * @return true if newly created, false otherwise (failed or namespace already present)
      */
     public static boolean createIfDoesNotExistsProject(String namespace) {
         OpenShift openShift = OpenShifts.master(namespace);
 
-        boolean isCreated = false;
+        // in case namespace is terminating (means is being deleted) then wait
```
mchoma, line 67: Sometimes it happen namespace is in terminating state forever. Motivation for introducing `xtf.openshift.namespace.per.testcase=true` was to mitigate forever terminating projects. So we did not avoid our initial motivation?

mchoma:
> I dislike all this hacked `xtf.openshift.namespace.per.testcase`

Initial motivation was to mitigate .clean() problems on namespaces which are Terminating. Possibly this can be extended to paralelism. So I think idea is good.

Problem is if this can be implemented in clear way. If we will have many corner cases in xtf `if OpenShiftConfig.useNamespacePerTestCase()` that would make xtf very fragile.

What exactly you find hacky about `xtf.openshift.namespace.per.testcase`?  Idea or implementation?

core/src/main/java/cz/xtf/core/namespace/NamespaceManager.java
mnovak1: if `xtf.openshift.namespace.per.testcase=true` then it's not an issue as it's creating namespace with different name.

junit5/src/main/java/cz/xtf/junit5/listeners/ProjectCreator.java
mnovak1: > Isn't it changing current behaviour?

Yes, it's intentional change and return to behavior in previous versions (XTF 0.26 and older). Deleting namespaces between test cases started to cause issues when `xtf.openshift.namespace.per.testcase=false` as there is no 100% reliable way to check that the namespace is deleted.

mnovak1: > Ok, so problem is we were deleting namespaces between testcases...

Yes.

>I agree we could use if xtf.openshift.namespace.per.testcase=true ( or deleteProjectIfUsedNamespacePerTestCase - it is the same).

 I've already updated the PR and moved problematic if statement to `NamespaceManager.deleteProjectIfUsedNamespacePerTestCase()` so usage of `OpenShiftConfig.useNamespacePerTestCase()` is isolated to this class.

mchoma: Ok. LGTM.

I think idea to be able to run one testsuite in own namespace is nice strategy. Seems it is hard to incorporate into existing XTF seamlessly.

mnovak1: Ok, resolve this thread.

mnovak1:
> > I dislike all this hacked `xtf.openshift.namespace.per.testcase`
> 
> Initial motivation was to mitigate .clean() problems on namespaces which are Terminating. Possibly this can be extended to paralelism. So I think idea is good.
> 
> Problem is if this can be implemented in clear way. If we will have many corner cases in xtf `if OpenShiftConfig.useNamespacePerTestCase()` that would make xtf very fragile.
> 
> What exactly you find hacky about `xtf.openshift.namespace.per.testcase`? Idea or implementation?

I believe this was discussed in other thread. Can this PR be Merged, please?

mchoma:
> I believe this was discussed in other thread. Can this PR be Merged, please?

Yes it can be merged. I am ok with it.

mchoma (approved):

mchoma:
@mnovak1 Just an idea. Maybe when we find we have a lot of `if (xtf.openshift.namespace.per.testcase=true)` in code and xtf code messy and fragile. We can think of separate those code branches by some design pattern; Command, Strategy, SPI, etc.

So for example have in one place `if xtf.openshift.namespace.per.testcase=true` and instantiate correct implementations which will be used, e.g. ProjectCreator or ProjectPerTestCaseCreator. Before writing any code, we would need to draw how that could work.

PR #512, opened 2022-09-21
Fixed permissions of shell scripts in binary builds

tommaso-borgato:
In binary builds from sources, when the source is an already provisioned server (e.g. some "target/server" produced by the `wildfly-maven-plugin`), then execution permissions on ".sh" files are lost;

This causes the following error when the POD starts:

```
/opt/jboss/container/wildfly/run/run: line 74: /opt/server/bin/openshift-launch.sh: Permission denied
```

This MR just fixes permissions on ".sh" files which is the bare minimum required to have a working server installation;

### DIFF
--- a/core/src/main/java/cz/xtf/core/bm/BinaryBuildFromSources.java
+++ b/core/src/main/java/cz/xtf/core/bm/BinaryBuildFromSources.java
@@ -10,6 +10,10 @@ import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.PosixFileAttributeView;
+import java.nio.file.attribute.PosixFileAttributes;
+import java.nio.file.attribute.PosixFilePermission;
 import java.security.DigestOutputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -17,6 +21,7 @@ import java.time.Instant;
 import java.util.Collection;
 import java.util.Date;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


mnovak1:
@tommaso-borgato I don't understand the details here, I just want to make sure that this is not workaround for something what is actually a bug. Shouldn't `wildfly-maven-plugin` preserve permission on `.sh` files?

mnovak1:
Or is the executable flag lost when creating tar?

tommaso-borgato:
@mnovak1 I am not an XTF expert, but I experienced the same issue on another project.... IIUC the java api used to assemble the tar (`TarArchiveEntry`) have issues with preserving permissions.... for our use case the fix I submitted is enough

fabiobrz:
> @mnovak1 the issue here is that classes in `org.apache.commons.compress.archivers.tar` e.g. `TarArchiveEntry` do not preserve permissions when building a `tar` .... this is actually a bug in `xtf`
> 
> it's no surprise to me that the issue is in the only step of the process which is using Java library for creating a `tar` because I experienced the very same issue on another project (the solution was running `zip` directly there - which we cannot apply here)
> 
> this MR is a workaround for the one scenario that we use: creating a tar stream containing a server directory to be sent to the builder image
> 
> a more comprehensive solution would be try and find and alternative to `org.apache.commons.compress.archivers.tar` which preserves permissions (if any) and re-write `BinaryBuildFromSources`
> 
> perhaps we might trace the more comprehensive solution in a separate issue (with lower priority smile )

From what I can read around, this might be intentional, since permissions are dealt with at the OS level.
In such a case, it seems to me this is not a workaround for an external component bug, but rather a proper fix for the XTF (caller) behavior when dealing with org.apache.commons.compress.archivers.tar.TarArchiveEntry etc.

In such a case there's no issue that should be created externally, maybe an internal one fixed by this very PR.
Let me know in case I am missing something.

tommaso-borgato:
@mnovak1 good idea, going to check

mnovak1:
> @mnovak1 good idea, going to check

:+1: ...thanks for looking at it.

tommaso-borgato:
@mnovak1 I applied your suggestion and the bug fix is now complete

fabiobrz (approved):
I am approving since this LGTM now. But leaving the last word to @mnovak1 of course :)


Maybe this could be extracted into some utils in the future and maybe we could add a comment identifying the source for such conversion.

mnovak1 (approved):

mnovak1:
LGTM, merging. Thanks @tommaso-borgato !

PR #522, opened 2022-12-15
Configure VolumeMounts with subPath in ContainerBuilder

tommaso-borgato:
This MR introduces the option to configure volumeMounts with subPath through cz.xtf.builder.builders.pod.ContainerBuilder

### DIFF
--- a/builder/src/main/java/cz/xtf/builder/builders/pod/ContainerBuilder.java
+++ b/builder/src/main/java/cz/xtf/builder/builders/pod/ContainerBuilder.java
@@ -128,6 +128,16 @@ public class ContainerBuilder implements EnvironmentConfiguration, ResourceLimit
         return this;
     }
 
+    public ContainerBuilder addVolumeMount(String name, String mountPath, boolean readOnly, String subPath) {
+        this.volumeMounts.add(new VolumeMount(name, mountPath, readOnly, subPath));
+        return this;
+    }
+
+    public ContainerBuilder addVolumeMount(VolumeMount volumeMount) {
+        this.volumeMounts.add(volumeMount);
+        return this;
+    }
+
     public LivenessProbe addLivenessProbe() {
         this.livenessProbe = new LivenessProbe();
         return (LivenessProbe) this.livenessProbe;
@@ -205,6 +215,7 @@ public class ContainerBuilder implements EnvironmentConfiguration, ResourceLimit
                 .withName(item.getName())
                 .withMountPath(item.getMountPath())
                 .withReadOnly(item.isReadOnly())
+                .withSubPath(item.getSubPath())
                 .build()).collect(Collectors.toList()));
 
         final List<ComputingResource> requests = computingResources.values().stream().filter(x -> x.getRequests() != null)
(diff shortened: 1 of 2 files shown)


fabiobrz (approved):
LGTM - @mnovak1 this seems conservative to me, since the only relevant change keeps the compatibility with the existing implementation. WDYT?

jbliznak (approved):

builder/src/main/java/cz/xtf/builder/builders/pod/ContainerBuilder.java
```diff
@@ -271,30 +277,6 @@ public void addPreStopHandler(Handler handler) {
         preStopHandler = handler;
     }
 
-    private static class VolumeMount {
```
simkam, line 274: I would keep changes minimal, it is used only in this class, I would keep it nested and not add any unnecessary changes. 

Definitely, making the new class public makes it part of public API which we don't want. Unless I'm missing something.

simkam (changes requested):

builder/src/main/java/cz/xtf/builder/builders/pod/ContainerBuilder.java
tommaso-borgato: unfortunately we need this class to be public in order to configure its `subPath` property, e.g.:
```
volumeMounts.add(
				new VolumeMount("data-dir", "/opt/server/standalone/data/ejb-xa-recovery", false, "ejb-xa-recovery"));
```

simkam: Can you please elaborate? I still don't see it. Isn't it
```
ContainerBuilder.addVolumeMount("data-dir", "/opt/server/standalone/data/ejb-xa-recovery", false, "ejb-xa-recovery")
```
?

tommaso-borgato: it's actually in another project

tommaso-borgato: I've sent you the link on the chat since it's an internal repo

tommaso-borgato: btw, in the snipped you pasted there a missing `new VolumeMount(...`

simkam: ok, looking at other code, now I understand ... but in this case, I would even extend it by one more signature 

```
public ContainerBuilder addVolumeMount(VolumeMount volumeMount) {
   this.volumeMounts.add(volumeMount);
   return this;
}
```

otherwise it doesn't make sense to me to make it part of public API because there is really no way to use it outside XTF.

tommaso-borgato: done! tnks for the suggestion

simkam (approved):

PR #523, opened 2023-02-09
add startup probe

### LINKED ISSUE #520: StartupProbes support
For some time there is third kind of probes in Kubernetes (since OCP 4.7) - StartupProbes. During the processing of EAP8 it was touched several times. 

It would be nice if we have such support in XTF as well, same as we do have `ReadinessProbe` and `LivenessProbe`. If we find ourselves in situation we would need it.

[1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

laDok8:

Adding startup probe functionality.

This code works as is, but if the startup probe fails tests still wait for a timeout. I propose also modifying waiter behavior to fail immediately if the startup probe is present and fails WDYT @mnovak1 ?

Resolves #520.

### DIFF
--- a/builder/src/main/java/cz/xtf/builder/db/AbstractDatabase.java
+++ b/builder/src/main/java/cz/xtf/builder/db/AbstractDatabase.java
@@ -24,6 +24,7 @@ public abstract class AbstractDatabase extends DefaultStatefulAuxiliary {
 
     protected boolean withLivenessProbe;
     protected boolean withReadinessProbe;
+    protected boolean withStartupProbe;
 
     public AbstractDatabase(String symbolicName, String dataDir) {
         this("testuser", "testpwd", "testdb", symbolicName, dataDir);
@@ -42,6 +43,16 @@ public abstract class AbstractDatabase extends DefaultStatefulAuxiliary {
         this.configureEnvironment = configureEnvironment;
     }
 
+    public AbstractDatabase(String symbolicName, String dataDir, boolean withLivenessProbe, boolean withReadinessProbe,
+            boolean withStartupProbe,
+            boolean configureEnvironment) {
+        this(symbolicName, dataDir);
+        this.withLivenessProbe = withLivenessProbe;
+        this.withReadinessProbe = withReadinessProbe;
+        this.withStartupProbe = withStartupProbe;
+        this.configureEnvironment = configureEnvironment;
+    }
+
     public AbstractDatabase(String symbolicName, String dataDir, PersistentVolumeClaim pvc) {
         this("testuser", "testpwd", "testdb", symbolicName, dataDir, pvc);
     }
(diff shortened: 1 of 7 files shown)


mnovak1 (approved):

mnovak1:
> > but if the startup probe fails tests still wait for a timeout. I propose also modifying waiter behavior to fail immediately if the startup probe is present and fails
> 

Do we have something similar for readiness probe? I'm just trying to understand how it would work.

laDok8:
> > but if the startup probe fails tests still wait for a timeout. I propose also modifying waiter behavior to fail immediately if the startup probe is present and fails
> 
> I'm not sure, doesn't openshift restart the container if probes fail? The second start might be successful and the test would continue

looking more into It you are correct, based on `restartPolicy` pods should restart

builder/src/main/java/cz/xtf/builder/db/AbstractDatabase.java
```diff
@@ -24,28 +24,42 @@ public abstract class AbstractDatabase extends DefaultStatefulAuxiliary {
 
     protected boolean withLivenessProbe;
     protected boolean withReadinessProbe;
+    protected boolean withStartupProbe;
 
     public AbstractDatabase(String symbolicName, String dataDir) {
         this("testuser", "testpwd", "testdb", symbolicName, dataDir);
     }
 
+    @Deprecated
```
mnovak1, line 33: I'm thinking whether it's necessary to deprecate methods without startup probes?

laDok8: It's not necessary, but I find it clearer to keep the number of constructors at a minimum eg.
```
public AbstractDatabase(String symbolicName, String dataDir, boolean withLivenessProbe, boolean withReadinessProbe)
public AbstractDatabase(String symbolicName, String dataDir, boolean withLivenessProbe, boolean withReadinessProbe, boolean configureEnvironment)
public AbstractDatabase(String symbolicName, String dataDir, boolean withLivenessProbe, boolean withReadinessProbe, boolean withStartupProbe, boolean configureEnvironment)
```
could IMO be easily misused

mnovak1: I see and agree that having less contructors is good. However we try to keep XTF backward compatible as much as possible. If there is no hard technical reason to deprecate them (for example in unprobable case when liveness probe would be removed from Openshift) we should keep them.

laDok8:
sure, here is the OCP run https://jenkins.eapqe.psi.redhat.com/job/eap-8.x-openshift-4-80-openjdk17/26/ @mnovak1

mnovak1 (approved):

mnovak1:
LGTM. Though I hope that "sudden" adding of startup probe for dabases will not cause any regressions. I think we should mention it in notes for next release. So  noting it for myself.

laDok8:
further note: problem may arise if the database doesn't start in: `(startupFailureThreshold * startupPeriodSeconds) + startupInitialDelaySeconds`.

PR #526, opened 2023-02-20
0.29 fix smongiar

smongiar:

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
@@ -2,10 +2,11 @@ package cz.xtf.core.openshift;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
+import java.util.Arrays;
 import java.util.Objects;
 
 import org.apache.commons.lang3.StringUtils;
@@ -56,9 +57,12 @@ class OpenShiftBinaryManager {
         } else {
             // If we are using an existing kubeconfig (or a default kubeconfig), we copy the original kubeconfig
             if (StringUtils.isNotEmpty(kubeconfig)) {
-                // We copy the specified kubeconfig
+                // flatten kubeconfig in case it contains certs/keys
                 try {
-                    Files.copy(Paths.get(kubeconfig), Paths.get(ocConfigPath), StandardCopyOption.REPLACE_EXISTING);
+                    Files.write(Paths.get(ocConfigPath),
+                            Arrays.asList(new OpenShiftBinary(OpenShifts.getBinaryPath(), null)
+                                    .execute("config", "view", "--kubeconfig", kubeconfig, "--flatten")),
+                            StandardCharsets.UTF_8);
                 } catch (IOException e) {
                     throw new RuntimeException(e);
                 }
@@ -67,7 +71,11 @@ class OpenShiftBinaryManager {
                 File defaultKubeConfig = Paths.get(getHomeDir(), ".kube", "config").toFile();
                 if (defaultKubeConfig.isFile()) {
                     try {
-                        Files.copy(defaultKubeConfig.toPath(), Paths.get(ocConfigPath), StandardCopyOption.REPLACE_EXISTING);
+                        Files.write(Paths.get(ocConfigPath),
+                                Arrays.asList(new OpenShiftBinary(OpenShifts.getBinaryPath(), null)
+                                        .execute("config", "view", "--kubeconfig", defaultKubeConfig.getAbsolutePath(),
+                                                "--flatten")),
+                                StandardCharsets.UTF_8);
                     } catch (IOException e) {
                         throw new RuntimeException(e);
                     }


mnovak1 (approved):
Could you just provide more context/motivation for this PR, please? I don't see into those authtencation methods. Thanks 

Otherwise code is LGTM.

smongiar:
Hi, @mnovak1 
It seems that connection through kubeconfig doesn't work when we use cerficates and private keys. 

One example is provided by:
`users:
- name: admin/cds149rd042m87juef2g
  user:
    client-certificate: admin.pem
    client-key: admin-key.pem
`
in the _kubeconfig file_

XTF framework actually doesn't allow connections with those key/certificate auth method

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
```diff
@@ -56,9 +58,22 @@ private OpenShiftBinary getBinary(String token, String username, String password
...
+                    Path targetDir = Paths.get(ocConfigPath).getParent();
+                    for (Path f : files) {
+                        String targetPath = targetDir + System.getProperty("file.separator") + f.getFileName();
+                        File target = new File(targetPath);
+                        if (!target.exists()) {
+                            Files.copy(f, Paths.get(targetDir.toString()).resolve(f.getFileName()));
+                        }
+                    }
+                    if (!Paths.get(ocConfigPath).toFile().exists()) {
+                        Files.copy(Paths.get(kubeconfig), Paths.get(ocConfigPath), StandardCopyOption.REPLACE_EXISTING);
```
mnovak1, line 74: Ok, maybe I might have suggestion for improvement. Not sure if it's valid. Can be `StandardCopyOption.REPLACE_EXISTING` moved to upper for loop when copying file and this `if` block removed.

mnovak1 (approved):

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
smongiar: Basically, for me, the goal was to copy the whole content of _$KUBECONFIG_ parent directory (_/home/user/.kube_) to the _ocConfigPath_ parent directory, with the purpose of having all the certificates, keys and everything coded into the _kubeconfig_ file

The "REPLACING_EXISTING" copy option was the main issue I faced, so I tried to avoid it :) 
It happened that, at time _"t0"_,  if the kubeconfig file "_oc.config_" is not present in the source directory (usually named "_config_" ex: _/home/user/.kube/config_ ) the _oc.config_ file would be deleted and overrides. 

So, I would like to check if a current file already exists. I want to avoid the deletion behaviour. Unfortunately, looking at _StandardCopyOption_, there are no doable solutions.  Let me know if it's clear to you.

simkam: I am confused by this code, can you please add some comments? 

Can we avoid mixing java.io and java.nio2 APIs in new code? 

`for` cycle copies all files, no? Why are 73-74 needed? The file should always be there after `for` cycle, no? (Maybe it is obvious, but I fail to see it)

mnovak1: Sorry, I'm still missing some pieces. I don't undestand why `config` file would not be present in source directory.

smongiar: @mnovak1 The `config` file is present in the source directory. But the XTF framework is looking for `oc.config` in the target. Ideally, since I would like to copy the whole `/home/user/.kube` directory, the effect of  `StandardCopyOption.REPLACE_EXISTING`, is to replace everything, deleting `oc.config` file in the target. Without that, the connection will fail.

@simkam Surely we can, I didn't tried with `java.nio2` APIs, but if you want I can work for the code update. 
About lines ` 73-74` and why the  `oc.config` would no more be there I already stated above. The difference between the previous logic, is that I want to copy whole directory content of `/home/user/.kube` instead of `/home/user/.kube/config` only

simkam: ok, I see ... the current execution flow is

1. create a new kubeconfig from a configured token or user name .... else
2. copy kubeconfig from the path specified in configuration `xtf.openshift.master.kubeconfig` ... else
3. copy default users kubeconfig from `~/.kube/config`

If we update 2 to copy all files in the directory, we should update also 3 to copy all files in the directory. 

But first, lines `63-72` do some weird `Path-String-File` conversion, please stick to nio only and `Path.get`, `Path.resolve` ...

------

Also what is the point of `FileSystems.getDefault()....` ?

-------

Also wouldn't `Files.walk...` be better readable. 

------

Also what is the point of 

```
                    if (!Paths.get(ocConfigPath).toFile().exists()) {
                        Files.copy(Paths.get(kubeconfig), Paths.get(ocConfigPath), StandardCopyOption.REPLACE_EXISTING);
```
? Wasn't the kubeconfig file copied already in cycle above?

simkam (changes requested):

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
mnovak1: > @mnovak1 The config file is present in the source directory. But the XTF framework is looking for oc.config in the target. Ideally, since I would like to copy the whole /home/user/.kube directory, the effect of StandardCopyOption.REPLACE_EXISTING, is to replace everything, deleting oc.config file in the target. Without that, the connection will fail.

Thanks, I think I've finally undestood it. So problem is that source directory contains file `config` which is copied to target directory but XTF is expecting that it will have name `oc.config`. So in the code above it's solved by copy `config` from source dir -> `oc.config` file in target dir.

It's not necessary as it's possible to change just `ocConfigPath` to point to the correct file in target directory like:
`ocConfigPath = targetDir.resolve(Paths.get(kubeconfig).getFileName()).toString();`

Then it's possible to get rid of `if` block and add `StandardCopyOption.REPLACE_EXISTING` to `for` loop to keep previous semantics that all in target dir is always overriden. 

I think this change and code clean up as suggested by @simkam should do the job.

smongiar: > If we update 2 to copy all files in the directory, we should update also 3 to copy all files in the directory.

That's not correct. Let me clarify better:

Because of 
https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java#L45

The "_oc.config_" file is created inside the _ocConfig folder_.

First, the purpose was to copy the content of the folder specified in "_xtf.openshift.master.kubeconfig_" property (i.e.: _/home/user/.kube/config_) in the _oc.config_ file

https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java#L61

The target directory would have been:

**Target:**
`
.
oc.config
`
Now, this behaviour doesn't take care of keys, certificates and other configurations things that are in the folder.

In the attempt to copy the whole content of folders instead of a single file, through
`Paths.get(kubeconfig).getParent()` ----> source directory 
`Paths.get(ocConfigPath).getParent()` ----> parent directory

The final result will be:

**Source:**
`
admin.pem
admin-key.pem
 .
config
`
**Target:**
`
admin.pem
admin-key.pem
.
config
`

That's because of _StandardCopyOption.REPLACE_EXISTING_ that replace everything and deletes the `oc.config` file. A solution could be to rename `oc.config` into `config` in https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java#L45, but I considered this too invasive.

wanting to keep the `oc.config` file without having to delete it, I make sure to copy only what doesn't exist inside the target directory:

https://github.com/smongiar/xtf/blob/0.29-fix-smongiar/core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java#L66-L71

smongiar: > I think this change and code clean up as suggested by @simkam should do the job
@mnovak1 sure, I will do that, the most important thing is the logic behind this PR is clear :)

mnovak1: > > I think this change and code clean up as suggested by @simkam should do the job
> 
> @mnovak1 sure, I will do that, the most important thing is the logic behind this PR is clear :)

@smongiar yes, at least we understand it now ;-)

simkam: @smongiar Thanks! Finally I got it.

But I don't like it, for example what if I set `xtf.openshift.master.kubeconfig=/home/user/Downloads/config` (that's something I sometimes really do when I download kubeconfig from Jenkins job), it would copy all files in Downloads. It should not assume that config is in configuration directory. 

Either we can try to assume a list of file names to copy or introduce new property/properties with additional final.

simkam: > Either we can try to assume a list of file names to copy or introduce new property/properties with additional files.

Or update the existing property to accept multiple files. I think it should be possible to update it and keep it backward compatible.

smongiar: @simkam 
I think we should take a decision about that.

Surely if you use the "Download" folder, copying everything is not a good practice, but it's just the use that you made :)

Moreover, adding two more properties for the key authentication, is back-compatible but we can't assume in the future, other files will be needed inside the `.kube` folder (take care about the fact that this folder is only for kubernetes/openshift purpose).

What do you suggest then? Can I proceed to add two more props?
In case I would add something like
`xtf.openshift.master.kubeconfig.admin-key`
`xtf.openshift.master.kubeconfig.admin-cert`
in https://github.com/smongiar/xtf/blob/0.29-fix-smongiar/core/src/main/java/cz/xtf/core/config/OpenShiftConfig.java

let me know if it's ok for you

mnovak1: Ok, let's discuss it. I'm more for adding new config properties like `xtf.openshift.master.kubeconfig.dir` and `xtf.openshift.master.kubeconfig.dir`, wdyt?

smongiar: @mnovak1 I think that specifying the whole directory as an option, we will fall into the same issues as before :)
we should manage the replacement, since the configuration file has, by default, always a different name (`oc.config`)

mnovak1: @smongiar By allowing new property for the whole kubeconfig dir, it will be expected that user does not mess it up by other unwanted files.

My idea is to keep current logic for `xtf.openshift.master.kubeconfig` and basically add another `if-then-else` block like:

if `xtf.openshift.master|admin.kubeconfig.dir` is set then

- copy whole content of source dir to target and assume (require) that `config` is name of file with configuration (it might be mixed with specifying `config` file in source dir by `xtf.openshift.master|admin.kubeconfig` property.

else if `kubeconfig` is set then

- keep current logic

else copy .kube dir to target dir (keeping current logi)

wdyt?

smongiar: >  and assume (require) that config is name of file with configuration

@mnovak1 if there's a chance to do that, and get rid of `oc.config` filename, it would be much better.
Do you know the reason behind the renaming of `oc.config` in place of `config` ? I don't want to break something about XTF framework

simkam: I'll let it to @mnovak1 , but I would like to avoid directory copy if possible.

Since XTF uses two users ("master" and "admin") I think we might add 4 new properties
```
xtf.openshift.master.kubeconfig # exists
xtf.openshift.master.master.key
xtf.openshift.master.master.cert
xtf.openshift.admin.kubeconfig # exists
xtf.openshift.master.admin.key
xtf.openshift.master.admin.cert
```

they could later be used even without kubeconfig in combination with `xtf.openshift.master.username` instead of password.

mnovak1: I've created issue https://github.com/xtf-cz/xtf/issues/527 to discuss it. Feel free to edit comment.

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
```diff
@@ -56,9 +59,13 @@ private OpenShiftBinary getBinary(String token, String username, String password
...
             if (StringUtils.isNotEmpty(kubeconfig)) {
-                // We copy the specified kubeconfig
+                // flatten kubeconfig in case it contains certs/keys
                 try {
-                    Files.copy(Paths.get(kubeconfig), Paths.get(ocConfigPath), StandardCopyOption.REPLACE_EXISTING);
+                    Files.deleteIfExists(Paths.get(ocConfigPath));
+                    Files.write(Paths.get(ocConfigPath),
+                            Arrays.asList(new OpenShiftBinary(OpenShifts.getBinaryPath(), null)
+                                    .execute("config", "view", "--kubeconfig", kubeconfig, "--flatten")),
+                            StandardCharsets.UTF_8, StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW);
```
simkam, line 68: can we flatten also default config bellow? line 77

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
```diff
@@ -111,4 +118,4 @@ private static String getHomeDir() {
     private Path getProjectOcConfigDir() {
         return Paths.get("tmp/oc/");
     }
-}
+}
```
simkam, line 121: please keep empty line

simkam (changes requested):
Thanks! We are looking into why CI is failing.

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
smongiar: done

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
smongiar: done

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
simkam, line 68:
```
Files.deleteIfExists(Paths.get(ocConfigPath));
```
can be removed as well as `StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW` options. Default behaviour seems enough

> If no options are present then this method works as if the [CREATE](https://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardOpenOption.html#CREATE), [TRUNCATE_EXISTING](https://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardOpenOption.html#TRUNCATE_EXISTING), and [WRITE](https://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardOpenOption.html#WRITE) options are present.

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
```diff
@@ -67,7 +74,10 @@ private OpenShiftBinary getBinary(String token, String username, String password
                 File defaultKubeConfig = Paths.get(getHomeDir(), ".kube", "config").toFile();
                 if (defaultKubeConfig.isFile()) {
                     try {
-                        Files.copy(defaultKubeConfig.toPath(), Paths.get(ocConfigPath), StandardCopyOption.REPLACE_EXISTING);
+                        Files.write(Paths.get(ocConfigPath),
+                                Arrays.asList(new OpenShiftBinary(OpenShifts.getBinaryPath(), null)
+                                        .execute("config", "view", "--kubeconfig", defaultKubeConfig.getAbsolutePath(), "--flatten")),
+                                StandardCharsets.UTF_8, StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW);
```
simkam, line 80: as above, `OpenOptions` can be omitted

simkam (changes requested):
two more comments. Sorry I didn't notice before :cry: . 

CI failures are due to https://github.com/xtf-cz/xtf/pull/531 ... please rebase

core/src/main/java/cz/xtf/core/openshift/OpenShiftBinaryManager.java
smongiar: done

smongiar:
> two more comments. Sorry I didn't notice before cry .
> 
> CI failures are due to #531 ... please rebase

rebased with the upstream master branch

simkam:
@smongiar did you push it?

smongiar:
> @smongiar did you push it?

I'm sorry, try again

simkam (approved):

mnovak1:
@smongiar Could you fix the formatting, please?  (`mvn clean install` should do the jobs)

Otherwise LGTM.

smongiar:
> @smongiar Could you fix the formatting, please? (`mvn clean install` should do the jobs)
> 
> Otherwise LGTM.

It should be ok, now

mnovak1:
Thanks, merging.

mnovak1:
And thanks for the fix! :-)

PR #528, opened 2023-02-23
Configure Apache compress - use POSIX due to big group ID number of linux user

kstekovi:
Due to big group ID number of the user the Apache compress throw exception when the user have big number as his group ID.  The big numbers can allowed by POSIX.


simkam (approved):

simkam:
I can see a small risk that it might not work on IBM ppc64 or s390, but I'm not sure we have a easy way to test it.

mnovak1:
LGTM

PR #532, opened 2023-03-10
Update Kubernetes Client to 6.X.X

RoyalKarma:
Adding compatibility with Kubernetes Client 6.X.X.

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
+++ b/core/src/main/java/cz/xtf/core/openshift/OpenShift.java
@@ -41,6 +41,7 @@ import cz.xtf.core.waiting.Waiter;
 import io.fabric8.kubernetes.api.builder.Visitor;
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.fabric8.kubernetes.api.model.Endpoints;
 import io.fabric8.kubernetes.api.model.EnvVar;
 import io.fabric8.kubernetes.api.model.EnvVarBuilder;
@@ -49,6 +50,7 @@ import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
 import io.fabric8.kubernetes.api.model.GenericKubernetesResourceList;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesList;
+import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.LocalObjectReferenceBuilder;
 import io.fabric8.kubernetes.api.model.Node;
 import io.fabric8.kubernetes.api.model.ObjectMetaBuilder;
@@ -70,11 +72,13 @@ import io.fabric8.kubernetes.api.model.rbac.RoleBinding;
 import io.fabric8.kubernetes.api.model.rbac.RoleBindingBuilder;
 import io.fabric8.kubernetes.api.model.rbac.Subject;
 import io.fabric8.kubernetes.api.model.rbac.SubjectBuilder;
-import io.fabric8.kubernetes.client.AppsAPIGroupClient;
 import io.fabric8.kubernetes.client.Config;
+import io.fabric8.kubernetes.client.KubernetesClientBuilder;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.LocalPortForward;
 import io.fabric8.kubernetes.client.dsl.LogWatch;
+import io.fabric8.kubernetes.client.dsl.base.ResourceDefinitionContext;
+import io.fabric8.kubernetes.client.dsl.internal.HasMetadataOperationsImpl;
 import io.fabric8.openshift.api.model.Build;
 import io.fabric8.openshift.api.model.BuildConfig;
 import io.fabric8.openshift.api.model.BuildRequest;
@@ -88,7 +92,8 @@ import io.fabric8.openshift.api.model.ProjectRequestBuilder;
 import io.fabric8.openshift.api.model.Route;
 import io.fabric8.openshift.api.model.RouteSpecBuilder;
 import io.fabric8.openshift.api.model.Template;
-import io.fabric8.openshift.client.DefaultOpenShiftClient;
+import io.fabric8.openshift.client.NamespacedOpenShiftClient;
+import io.fabric8.openshift.client.NamespacedOpenShiftClientAdapter;
 import io.fabric8.openshift.client.OpenShiftConfig;
 import io.fabric8.openshift.client.OpenShiftConfigBuilder;
 import io.fabric8.openshift.client.ParameterValue;
@@ -97,7 +102,7 @@ import rx.Observable;
 import rx.observables.StringObservable;
 
 @Slf4j
-public class OpenShift extends DefaultOpenShiftClient {
+public class OpenShift extends NamespacedOpenShiftClientAdapter {
     private static ServiceLoader<CustomResourceDefinitionContextProvider> crdContextProviderLoader;
     private static volatile String routeSuffix;
 
@@ -115,7 +120,6 @@ public class OpenShift extends DefaultOpenShiftClient {
      * NOTE: at the moment only place where this is used is for labeling namespaces. Other usages may be added in the future.
      */
     public static final String XTF_MANAGED_LABEL = "xtf.cz/managed";
-    private final AppsAPIGroupClient appsAPIGroupClient;
 
     /**
      * Autoconfigures the client with the default fabric8 client rules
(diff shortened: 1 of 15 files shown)


core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -277,15 +279,20 @@ public KubernetesList createResources(List<HasMetadata> resources) {
     }
 
     public KubernetesList createResources(KubernetesList resources) {
-        return lists().create(resources);
+        return (KubernetesList) resourceList(resources.getItems()).create();
     }
 
     public boolean deleteResources(KubernetesList resources) {
-        return lists().delete(resources);
+        if (resourceList(resources.getItems()).delete() != null) {
```
simkam, line 286: I think that `resourceList(resources.getItems()).delete()` will always be non null. I think we can compare size of original list with delete list 

```
return resources.getItems().size() == resourceList(resources.getItems()).delete().size(); 
```

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -277,15 +279,20 @@ public KubernetesList createResources(List<HasMetadata> resources) {
...
+            return true;
+        } else {
+            return false;
+        }
     }
 
     public void loadResource(InputStream is) {
-        load(is).deletingExisting().createOrReplace();
+        load(is).withTimeout(30, TimeUnit.SECONDS).delete();
```
simkam, line 294: do we need timeout? I know it is in example in migration guide, but I would let it use default.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -366,12 +373,12 @@ public boolean deleteProject() {
     }
 
     public boolean deleteProject(String name) {
-        return getProject(name) != null ? projects().withName(name).delete() : false;
+        return getProject(name) != null ? projects().withName(name).delete() != null : false;
```
simkam, line 376: the same as above with delete

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -396,12 +403,16 @@ public List<StatefulSet> getStatefulSets() {
     }
 
     public boolean deleteImageStream(ImageStream imageStream) {
-        return imageStreams().delete(imageStream);
+        if (imageStreams().resource(imageStream).delete() != null) {
```
simkam, line 406: the same as above with delete

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -413,12 +424,17 @@ public List<ImageStreamTag> getImageStreamTags() {
     }
 
     public boolean deleteImageStreamTag(ImageStreamTag imageStreamTag) {
-        return imageStreamTags().delete(imageStreamTag);
+
```
simkam, line 427: the same as above with delete

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -560,7 +576,11 @@ public boolean deletePod(Pod pod) {
     }
 
     public boolean deletePod(Pod pod, long gracePeriod) {
-        return pods().withName(pod.getMetadata().getName()).withGracePeriod(gracePeriod).delete();
+        if (pods().withName(pod.getMetadata().getName()).withGracePeriod(gracePeriod).delete() != null) {
```
simkam, line 579: the same as above with delete

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -571,11 +591,20 @@ public boolean deletePod(Pod pod, long gracePeriod) {
      * @return True if any pod has been deleted
      */
     public boolean deletePods(String key, String value) {
-        return pods().withLabel(key, value).delete();
+        if (pods().withLabel(key, value).delete() != null) {
```
simkam, line 594: the same as above with delete

test-helpers/src/main/java/cz/xtf/testhelpers/image/ImageMetadata.java
```diff
@@ -51,8 +51,7 @@ public static ImageMetadata prepare(OpenShift openShift, Image image) {
                 .withName(image.getRepo() + ":" + image.getMajorTag()).get();
         Waiter metadataWaiter = new SimpleWaiter(() -> {
             ImageStreamTag isTag = imageStreamTagSupplier.get();
-            if (isTag != null && isTag.getImage() != null && isTag.getImage().getDockerImageMetadata() != null
-                    && isTag.getImage().getDockerImageMetadata().getAdditionalProperties() != null) {
+            if (isTag != null && isTag.getImage() != null&& isTag.getImage().getMetadata() !=null && isTag.getImage().getMetadata().getAdditionalProperties() != null) {
```
simkam, line 54: keep line break

simkam (changes requested):
comments in code, please fix all deletes in `OpenShift.java`, I commented only few of them

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
simkam: I think we can assume it would be just one resource, so either 
```
imageStreams().resource(imageStream).delete().size() == 1
```
or
```
!imageStreams().resource(imageStream).delete().size().isEmpty()
```

simkam (approved):

mnovak1:
@RoyalKarma  thanks, we will need CI run with EAP Openshift TS before this gets merged.

core/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<parent>
 		<artifactId>xtf-parent</artifactId>
 		<groupId>cz.xtf</groupId>
-		<version>0.29-SNAPSHOT</version>
+		<version>0.29-kubernetes-client-upgrade-SNAPSHOT</version>
```
simkam, line 8: ah, sorry, I didn't get you mean this version. Please revert it back in PR.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -389,15 +385,15 @@ public List<ImageStream> getImageStreams() {
 
     // StatefulSets
     public StatefulSet createStatefulSet(StatefulSet statefulSet) {
-        return appsAPIGroupClient.statefulSets().resource(statefulSet).create();
+        return OpenShift.get(getConfiguration()).apps().statefulSets().resource(statefulSet).create();
```
simkam, line 388: just `apps().statefulSets().resource(statefulSet).create()` doesn't work? This already runs in the context of `Openshift` instance, no need to create a new instance.

simkam (changes requested):

pom.xml
```diff
@@ -61,7 +61,7 @@
 
         <!-- Dependency version properties -->
         <version.httpclient>4.5.13</version.httpclient>
-        <version.openshift-client>5.12.2</version.openshift-client>
+        <version.openshift-client>6.4.1</version.openshift-client>
```
mnovak1, line 64: Latest version is `6.7.2`. Would it be possible to update to this version?

mnovak1 (commented):
LGTM, only question to update to latest version.

pom.xml
RoyalKarma: Should be possible, but there were some changes to the client again. I'll see if I can get it to work with them.

core/src/main/java/cz/xtf/core/openshift/OpenShift.java
```diff
@@ -88,16 +88,16 @@
...
 import io.fabric8.openshift.client.OpenShiftConfigBuilder;
 import io.fabric8.openshift.client.ParameterValue;
+import io.fabric8.openshift.client.impl.OpenShiftClientImpl;
 import lombok.extern.slf4j.Slf4j;
 import rx.Observable;
 import rx.observables.StringObservable;
 
 @Slf4j
-public class OpenShift extends DefaultOpenShiftClient {
+public class OpenShift extends OpenShiftClientImpl {
```
fabiobrz, line 100: Are we sure we want to extend on an `*Impl` class?

RoyalKarma: You are right, I've noticed the issue when I tried to use 6.7.2 version of the client as Mirek suggested. I'm going to have to come up with something else that hopefuly doesnt break all the other changes.

mnovak1: > .. Went for replacing inheritance with composition 

Sounds good to me.

fabiobrz (commented):
This seems to be [working for Intersmash](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/Intersmash_intersmash/69/pull-ci-Intersmash-intersmash-main-intersmash-e2e/1696904479976198144) but please notice that projects depending on strimzi for Kafka would have issues since their latest release (0.36.1) [supports Fabric8 Kubernetes client v. 6.7.2](https://github.com/strimzi/strimzi-kafka-operator/blob/0.36.1/pom.xml#L124-L126), which is not compatible with 6.8.1 (planned for 0.37).

RoyalKarma:
TS run after update to 6.8.1
https://jenkins.eapqe.psi.redhat.com/job/eap-8.x-openshift-ts-face/65/

simkam (approved):
please squash commits

PR #533, opened 2023-03-13
[bulder] Add MsSQL and OracleXE

OndrejGerzicak:
This PR adds OracleXe and MsSQL Server databases to the builder. Since the provided images don't create custom databases and users on setup, we need to use the default, that's why I added the new constructors to AbstractDatabase and AbstractSQLDatabase classes.
Because the defualt database of OracleXE is `XE` I've come across a possible issue with Uppercase dbName that gets to the opensihft deployment config name. Although the OracleXE database server is case insensitive, it is possible to crate a case sensitive SQL server so it's seems safer to me to use lower case equivalent in the deployment name deployment config name


builder/src/main/java/cz/xtf/builder/db/MsSQL.java
```diff
@@ -0,0 +1,114 @@
...
+        super.configureContainer(containerBuilder);
+        
+        //Setting up the anyuid service account
+        try (OpenShift master = OpenShifts.master()) {
+            if (master.getServiceAccount(ANYUID_SERVICE_ACCOUNT) == null) {
+                master.createServiceAccount(
+                        new ServiceAccountBuilder().withNewMetadata().withName(ANYUID_SERVICE_ACCOUNT).endMetadata().build());
+            }
+        }
+        try (OpenShift admin = OpenShifts.admin()) {
```
mchoma, line 62: This is common to both MsSQL and OracleXE. Have you tried to  avoid using `admin` ? It can happen  we will be forced to run this code on cluster without `admin`. Would that be possible?  If not can you document which permission regular `master` user will be missing?

builder/src/main/java/cz/xtf/builder/db/OracleXE.java
```diff
@@ -0,0 +1,74 @@
...
+
+import cz.xtf.builder.builders.pod.PersistentVolumeClaim;
+import cz.xtf.core.image.Image;
+
+/**
+ * Note that to be able to use the oracleXE on Openshift, you'll probably have to run it as user with uid 54321.
+ * For that you'll need to add anyuid service account
+ *
+ * * Also, the OracleXE image doesn't provide any option to add custom database and user on setup.
+ */
```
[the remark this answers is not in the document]
mchoma: > setting up the service account in this class, or at least set up the container to use the needed uid and pass the service account name to constructors

Yes, I would expect this class take care of everything what needs to be done for starting database. And what you mention sounds as prerequisite of that. So yes, me personally would expect to happen it here in this class.

builder/src/main/java/cz/xtf/builder/db/MsSQL.java
OndrejGerzicak: Regular master is missing permissions to add security context constraint. I added this info in the comment before the classes.

builder/src/main/java/cz/xtf/builder/db/MsSQL.java
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * By using this class you are accepting the Microsoft's EULA: https://go.microsoft.com/fwlink/?linkid=857698
+ */
+public class MsSQL extends AbstractSQLDatabase {
+
+    private final String ANYUID_SERVICE_ACCOUNT = "anyuid";
+    private final long USER_ID = 0;
+
+    public MsSQL() {
+        super("sa", "myMsSQLPassword123", "master", "MSSQL", "/var/lib/mssql/data");
```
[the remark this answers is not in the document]
OndrejGerzicak: Done

builder/src/main/java/cz/xtf/builder/db/OracleXE.java
```diff
@@ -0,0 +1,117 @@
...
+        super(USERNAME, PASSWORD, DB_NAME, symbolicName, DATA_DIR, withLivenessProbe, withReadinessProbe);
+    }
+
+    @Override
+    protected void configureContainer(ContainerBuilder containerBuilder) {
+        super.configureContainer(containerBuilder);
+
+        //Setting up the anyuid service account
+        try (OpenShift master = OpenShifts.master()) {
+            if (master.getServiceAccount("ANYUID_SERVICE_ACCOUNT") == null) {
```
jbliznak, line 63: this needs to be unquoted to actually use the variable's value

builder/src/main/java/cz/xtf/builder/db/MsSQL.java
```diff
@@ -0,0 +1,120 @@
...
+    public MsSQL(String symbolicName, boolean withLivenessProbe, boolean withReadinessProbe) {
+        super(USERNAME, PASSWORD, DB_NAME, symbolicName, DATA_DIR, withLivenessProbe, withReadinessProbe);
+    }
+
+    @Override
+    protected void configureContainer(ContainerBuilder containerBuilder) {
+        super.configureContainer(containerBuilder);
+        
+        //Setting up the anyuid service account
+        //Don't close the OpenShift Client instance, it could cause trouble with pod shell access
```
jbliznak, line 63: Not sure we want this info here, it is a general note more suited for README or maybe `OpenShifts.master()` javadoc .. but it isn't a socpe of this PR at all

OndrejGerzicak: I added this comment just in case someone else would see this line and wanted to use the `try with resource` block as it's suggested by IDEA.
Agree that it should be mentioned elsewhere though.

jbliznak (approved):

builder/src/main/java/cz/xtf/builder/db/MsSQL.java
mchoma: If IDEA is suggesting something we know can bring problem, can it be fixed in XTF code? Can we handle it as XTF bug?

mchoma (approved):

builder/src/main/java/cz/xtf/builder/db/MsSQL.java
jbliznak: you can skip the IDEA from suggesting this by introducing comment
```suggestion
        //noinspection resource
```

But IDEA is not the only IDE in the world so I don't know if we want to go this road

mnovak1 (approved):

mnovak1:
LGTM. merging. Thanks!

PR #535, opened 2023-04-03
Downgrade impsort plugin to JDK 8 compatible version

mnovak1:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -299,6 +299,24 @@
                 </executions>
             </plugin>
 
+            <plugin>
+                <groupId>net.revelc.code</groupId>
+                <artifactId>impsort-maven-plugin</artifactId>
+                <version>1.6.2</version>
+                <configuration>
+                    <groups>java.,javax.,org.,com.</groups>
+                    <staticGroups>*</staticGroups>
+                    <removeUnused>true</removeUnused>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>sort-imports</id>
+                        <goals>
+                            <goal>sort</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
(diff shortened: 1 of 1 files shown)


simkam:
So we are dropping support for JDK 8?

mnovak1:
> So we are dropping support for JDK 8?

Nope. This is to unblock this check in GH action. All artificts has still compile target to JDK 8.

simkam:
```
                <configuration>
                    <source>${maven.compiler.source}</source>
                    <target>${maven.compiler.target}</target>
                </configuration>
```
isn't enough. We should then switch to using `--release` - https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

mnovak1:
> ```
>                 <configuration>
>                     <source>${maven.compiler.source}</source>
>                     <target>${maven.compiler.target}</target>
>                 </configuration>
> ```
> 
> isn't enough. We should then switch to using `--release` - https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

I'm not sure I understand. Could you elaborate more, please?

simkam:
Unlike `source` and `target`, `release` ensures that only bytecode usable by the selected version is produced by the build. More details for example in https://www.morling.dev/blog/bytebuffer-and-the-dreaded-nosuchmethoderror/ .

mnovak1:
Thanks. I've updated the compiler plugin to set release version as well.

mnovak1:
@simkam I've lowered version of impsort plugin to be JDK 8 compatible. However it's just temporary solution. I would like to migrate to Spotless in near future.

simkam:
ok, but please update commit message and PR description

mnovak1:
done

PR #537, opened 2023-04-11
Fixing BuildManager in order not to create a system:image-puller role for the authenticated users when the master and build namespaces are set to the same value

### LINKED ISSUE #536: Is BuildManager wrongly creating a system:image-puller RoleBinding?
When being instantiated, `BuildManager` wants for the authenticated users group to be able and access images it stores, see [describe rolebinding.rbac](https://docs.openshift.com/container-platform/4.12/openshift_images/managing_images/using-image-pull-secrets.html#images-allow-pods-to-reference-images-across-projects_using-image-pull-secrets)

This is described by the docs, see https://docs.openshift.com/container-platform/4.12/openshift_images/managing_images/using-image-pull-secrets.html#images-allow-pods-to-reference-images-across-projects_using-image-pull-secrets but in some cases it clashes with downstream business logic, e.g.: when calling OpenShiftWaiters.isProjectClean(), a list of removable resources is collected and the an unexpected role binding - named `system:image-puller` is found, holding the connection between the `system:image-puller` _role_ itself and the authenticated users group.

This makes tests that assert `isProjectClean()` fail unexpectedly.

**Note**: it has been suggested that the issue wouldn't be hit in case the `xtf.openshift.namespace` and the `xtf.bm.namespace` properties are set to two different values. In such a case the case when those are not should be handled.

fabiobrz:
Fix #536 

Tests passed locally.

---

### DIFF
--- a/core/src/main/java/cz/xtf/core/bm/BuildManager.java
+++ b/core/src/main/java/cz/xtf/core/bm/BuildManager.java
@@ -49,7 +49,9 @@ public class BuildManager {
             openShift.setupPullSecret(OpenShiftConfig.pullSecret());
         }
 
-        openShift.addRoleToGroup("system:image-puller", "ClusterRole", "system:authenticated");
+        if (!BuildManagerConfig.namespace().equals(OpenShiftConfig.namespace())) {
+            openShift.addRoleToGroup("system:image-puller", "ClusterRole", "system:authenticated");
+        }
     }
 
     public ManagedBuildReference deploy(ManagedBuild managedBuild) {


mnovak1 (commented):
LGTM, however we'll need some comparing CI runs with same and different namespaces.

fabiobrz:
> LGTM, however we'll need some comparing CI runs with same and different namespaces.

Thanks @mnovak1 - I'll try to collect and share some link to upstream runs in here.

mnovak1:
@fabiobrz do we have a passing ci run?

fabiobrz:
> @fabiobrz do we have a passing ci run?

HI @mnovak1 - I released a XTF snapshot containing the fix, and submitted a couple of Intersmash PRs to check it, one uses the same name for both the namespaces [1], and one is changing the one for the build namespace [2].
Let me know whether anything else is needed. 

[1]
https://github.com/Intersmash/intersmash/pull/44

[2]
https://github.com/Intersmash/intersmash/pull/45

mnovak1 (approved):
LGTM, merging

PR #538, opened 2023-04-11
Adding a validation method to XTFConfig to check that build and master namespace don't collide

### LINKED ISSUE #536: Is BuildManager wrongly creating a system:image-puller RoleBinding?

fabiobrz:
Adding a validation method to XTFConfig, which is called by a static block in order to check that build and master namespace don't collide since this might cause leftover resources to be found by OpenShift::clean()

Fix #536 - alternative to #537 

### DIFF
--- a/core/src/main/java/cz/xtf/core/config/XTFConfig.java
+++ b/core/src/main/java/cz/xtf/core/config/XTFConfig.java
@@ -35,6 +35,7 @@ public final class XTFConfig {
     // Pre-loading
     static {
         loadConfig();
+        validateConfig();
     }
 
     public static void loadConfig() {
@@ -49,6 +50,13 @@ public final class XTFConfig {
 
         // Set new values based on old properties if new are not set
         BackwardCompatibility.updateProperties();
+    }
+
+    public static void validateConfig() {
+        if (BuildManagerConfig.namespace().equals(OpenShiftConfig.namespace())) {
+            throw new IllegalStateException(
+                    "The xtf.openshift.namespace and xtf.bm.namespace properties cannot be set to the same value.");
+        }
     }
 
     public static String get(String property) {


simkam:
I'm not sure whether we should only print a warning. I guess in some cases it might work without issues even when namespace is the same.

mnovak1:
@fabiobrz AFACT we allow to set build namespace to master namespace. For example we use it for CVP testing. Thus we'll need to solve all issues instead.

fabiobrz:
Thanks @simkam, indeed it would work in some cases, while it would not in others, as the one I described in the PR description and [here](https://github.com/xtf-cz/xtf/issues/536)

Thanks @mnovak1 - I'll put this one back to draft, then.

mnovak1:
@fabiobrz I'll close this PR as we'll need to resolve issues with the same build and master namespaces instead of trying to avoid it.

PR #540, opened 2023-04-20
Removed HELM_KUBEASUSER from Helm binary + changed rvelc plugin versions

spriadka:
I have found out that combining `HELM_KUBEASUSER` and `KUBETOKEN` does not work well with `masterBinary`. Removing `HELM_KUBEASUSER` seems to solve the issue. In addition to this change, I've introduced changes needed to compile the project on Maven `3.9.1` with JDK 8 & 11.


mnovak1:
Yes, closing this one.

PR #544, opened 2023-04-27
Fix impsort mvn 391

mnovak1:
plexus is no longer injected by maven

https://maven.apache.org/docs/3.9.1/release-notes.html

> Maven 2.x was auto-injecting an ancient version of plexus-utils dependency into the plugin classpath, and Maven 3.x continued doing this to preserve backward compatibility. Starting with Maven 3.9, it does not happen anymore. This change may lead to plugin breakage. The fix for affected plugin maintainers is to explicitly declare a dependency on plexus-utils. The workaround for affected plugin users is to add this dependency to plugin dependencies until issue is fixed by the affected plugin maintainer. See [MNG-6965](https://issues.apache.org/jira/browse/MNG-6965).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -303,6 +303,13 @@
                 <groupId>net.revelc.code</groupId>
                 <artifactId>impsort-maven-plugin</artifactId>
                 <version>1.6.2</version>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.codehaus.plexus</groupId>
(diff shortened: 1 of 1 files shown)


simkam (approved):

mnovak1:
@simkam I've force pushed the PR as I left a mess there.

simkam:
ha, I thought it is intentional ... ok

Issue #573, opened 2024-08-19
/opt/mssql-tools/bin/sqlcmd: not found in mcr.microsoft.com/mssql/server:2022-latest


mchoma:
Latest mssql (2022-latest refering to 2022-CU14-ubuntu-20.04) is not starting properly because we are getting probe error

```
Readiness probe failed: /bin/sh: 0: can't access tty; job control turned off /bin/sh: 1: /opt/mssql-tools/bin/sqlcmd: not found 
```

Commadn wasreplaced `/opt/mssql-tools/bin/sqlcmd` -> `/opt/mssql-tools18/bin/sqlcmd`. From tracking issue  https://github.com/microsoft/mssql-docker/issues/892, it does not seem adding symlink is plan. 

Seems to me we need to change this code in xtf https://github.com/xtf-cz/xtf/blob/1daa976b2e32c0bc9fcdfda1ae947d37836c15fa/builder/src/main/java/cz/xtf/builder/db/MsSQL.java#L92-L104 to be able to use latest mssql images again.

Also from issue comments we may need to add option `-C`

mchoma:
@dosu any idea?

PR #575, opened 2024-08-26
sqlcmd path updated for the newest MsSQL image

dcihak:
In the latest MSSQL Docker image (2022-latest refering to 2022-CU14-ubuntu-20.04) the path to sqlcmd was changed. This causes probe error during the start.

Tracked by [the mssql-docker issue](https://github.com/microsoft/mssql-docker/issues/892).

Related XTF issue https://github.com/xtf-cz/xtf/issues/573.

### DIFF
--- a/builder/src/main/java/cz/xtf/builder/db/MsSQL.java
+++ b/builder/src/main/java/cz/xtf/builder/db/MsSQL.java
@@ -94,11 +94,11 @@ public class MsSQL extends AbstractSQLDatabase {
         return new ProbeSettings(30,
                 String.valueOf(this.getPort()),
                 5,
-                "/opt/mssql-tools/bin/sqlcmd -S localhost -d " + getDbName() + " -U " + getUsername() + " -P " + getPassword()
-                        + " -Q \"Select 1\"",
+                "/opt/mssql-tools18/bin/sqlcmd -S localhost -d " + getDbName() + " -U " + getUsername() + " -P " + getPassword()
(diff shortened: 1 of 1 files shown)


simkam (approved):
Documented directly in https://hub.docker.com/r/microsoft/mssql-server

> Starting with SQL Server 2022 CU 14, we are updating SQL Server 2022 container images to include the new mssql-tools18 package. With the introduction of SQL Server 2022 CU 14, and in all future container images, the previous directory /opt/mssql-tools/bin will be phased out. The new directory for Microsoft ODBC 18 tools will be '/opt/mssql-tools18/bin', aligning with the latest tools offering. For additional details on all the changes and security enhancements, please refer to the blog post [ODBC Driver 18.0 for SQL Server Released - Microsoft Community Hub⁠](https://techcommunity.microsoft.com/t5/sql-server-blog/odbc-driver-18-0-for-sql-server-released/ba-p/3169228). This new iteration, ODBC driver version 18, is designed with an ‘encryption-first’ approach, ensuring that utilities like sqlcmd and bcp that utilize the Microsoft ODBC driver will operate under the ‘secure by default’ principle. Users who wish to disable encryption will need to do so explicitly.

I don't see a need to stay compatible with older image versions.

mnovak1 (approved):
Merging even though this might cause some issues in case users are still on older hardcoded version of mssql image. They need to either update the mssql image or override the probe settings to old location.

Issue #572, opened 2024-08-16
mcr.microsoft.com/mssql/rhel/server:2022-CU13-rhel-9.1 should have one or two '/' 


mchoma:
XTF does not allow for me to use image like `mcr.microsoft.com/mssql/rhel/server:2022-CU13-rhel-9.1`

```
java.lang.IllegalArgumentException: image 'mcr.microsoft.com/mssql/rhel/server:2022-CU13-rhel-9.1' should have one or two '/' characters
	at cz.xtf.core.image.Image.from(Image.java:71)
	at cz.xtf.core.image.Image.get(Image.java:18)
	at cz.xtf.core.image.Image.resolve(Image.java:22)
	at cz.xtf.builder.db.MsSQL.getImageName(MsSQL.java:113)
	at cz.xtf.builder.db.AbstractDatabase.configureDeployment(AbstractDatabase.java:186)
	at cz.xtf.builder.db.AbstractDatabase.configureDeployment(AbstractDatabase.java:181)
```

We need `case 4+: ` section here? 

https://github.com/xtf-cz/xtf/blob/1daa976b2e32c0bc9fcdfda1ae947d37836c15fa/core/src/main/java/cz/xtf/core/image/Image.java#L54-L71

So we will end up with something like this
```
registry = mcr.microsoft.com
user = mssql/rhel
repoTag = server:2022-CU13-rhel-9.1
```

mchoma:
@dosu what do you think?

mchoma:
@dosu any objection with linked pull request? Do you think it address original issue? Are linked tests correctly testing change?

mchoma:
@dosu No that check is not implemented in https://github.com/xtf-cz/xtf/pull/577/

PR #577, opened 2024-08-30
[#572] Allow image url with more than 2 slashes to be parsed properly.

mnovak1:

CI runs:  
Without fix: https://url.corp.redhat.com/a7db151
With fix: https://url.corp.redhat.com/bbed950
* GalleonDatasourceNoDriverVersionTest#noMsSQLDriverVersion,GalleonDatasourceFeaturePackMsSQLTest - caused by not updated mssql image
* Rerun HAServletCounterUpstreamKubePingTest,RollingUpdatesKubePingTest → https://url.corp.redhat.com/e8c1138 passed

### DIFF
--- a/core/src/main/java/cz/xtf/core/image/Image.java
+++ b/core/src/main/java/cz/xtf/core/image/Image.java
@@ -1,6 +1,7 @@
 package cz.xtf.core.image;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import cz.xtf.core.config.XTFConfig;
@@ -70,7 +71,10 @@ public class Image {
                 repoTag = slashTokens[2];
                 break;
             default:
-                throw new IllegalArgumentException("image '" + imageUrl + "' should have one or two '/' characters");
+                registry = slashTokens[0];
+                user = slashTokens[1];
+                repoTag = String.join("/", Arrays.copyOfRange(slashTokens, 2, slashTokens.length));
+                break;
         }
 
         final String[] tokens = repoTag.split(":");
--- a/core/src/test/java/cz/xtf/core/image/ImageTest.java
+++ b/core/src/test/java/cz/xtf/core/image/ImageTest.java
@@ -99,4 +99,50 @@ public class ImageTest {
         Assertions.assertThrows(UnknownImageException.class, () -> Image.resolve("blabla"),
                 "Expected UnknownImageException to be thrown for unknown image.");
     }
+
+    @Test
+    public void testImageFromUrlOneToken() {
+        Image image = Image.from("nginx");
+        Assertions.assertTrue(image.getRegistry().isEmpty(), "Registry for image url: 'nginx' must be empty.");
+        Assertions.assertTrue(image.getUser().isEmpty(), "User for image url: 'nginx' must be empty.");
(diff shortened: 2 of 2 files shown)


core/src/test/java/cz/xtf/core/image/ImageTest.java
```diff
@@ -99,4 +99,50 @@ public void testNotResolvableImage() {
...
+        Assertions.assertEquals("nginx", image.getRepo(), "Wrong repository name.");
+        Assertions.assertTrue(image.getTag().isEmpty(), "Tag for image url: 'nginx' must be empty.");
+    }
+
+    @Test
+    public void testImageFromUrlTwoTokens() {
+        Image image = Image.from("user/nginx");
+        Assertions.assertTrue(image.getRegistry().isEmpty(), "Registry for image url: 'user/nginx' must be empty.");
+        Assertions.assertEquals(image.getUser(), "user", "User for image url: 'user/nginx' is wrong.");
+        Assertions.assertEquals("nginx", image.getRepo(), "Wrong repository name.");
```
simkam, line 117: weird order of expected/actual

> [assertEquals](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertEquals-byte-byte-java.lang.String-)(byte expected, byte actual, [String](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html?is-external=true) message)

below occurrences as well

simkam (changes requested):

core/src/test/java/cz/xtf/core/image/ImageTest.java
mnovak1: Thanks, fixed.

simkam (approved):

core/src/main/java/cz/xtf/core/image/Image.java
```diff
@@ -70,7 +71,10 @@ public static Image from(String imageUrl) {
                 repoTag = slashTokens[2];
                 break;
             default:
-                throw new IllegalArgumentException("image '" + imageUrl + "' should have one or two '/' characters");
+                registry = slashTokens[0];
+                user = slashTokens[1];
```
mchoma, line 75: This is different to what dosu says :) My understanding is aligned with dosu implementation.

IIUC dosu implementation https://github.com/xtf-cz/xtf/issues/572#issuecomment-2296478769 will bring this result
```
registry = mcr.microsoft.com
user = mssql/rhel
repoTag = server:2022-CU13-rhel-9.1
```

whereas your implementation will bring 

```
registry = mcr.microsoft.com
user = mssql
repoTag = rhel/server:2022-CU13-rhel-9.1
```

simkam: to me the latter makes more sense, though we are only bending image url (which can be almost anything) to what the class expects

mchoma: Yes, you was faster then me wanted to write `user = mssql ` makes more sense to me now while thinking about it.

mchoma (approved):
I consider https://github.com/xtf-cz/xtf/issues/572 for addressed. Now it is enabled to use image url with more than 2 slashes.

Tests proove results are expected.

Issue #581, opened 2024-09-17
CLIUtils creates a new thread pool for every command and does not terminate a threads


kstekovi:
When is used `SimpleWaiter` together this OpenShift binary it create two threads for every check of current state of resource in the OpenShift.  So it can create hundreds of threads when it waits for start of a POD. Unfortunately these thread are never terminated. There are sleeping or waiting. 

![ThreadPool](https://github.com/user-attachments/assets/d12f6519-7221-4357-af38-4c9ba34b3728)

simkam:
resolved by https://github.com/xtf-cz/xtf/pull/582

PR #582, opened 2024-09-17
[#581] close Output and Error stream threads after command is done

kstekovi:
…command

### DIFF
--- a/core/src/main/java/cz/xtf/core/openshift/CLIUtils.java
+++ b/core/src/main/java/cz/xtf/core/openshift/CLIUtils.java
@@ -44,6 +44,8 @@ public class CLIUtils {
             });
 
             int result = p.waitFor();
+            // Shutdown the output and error streams
+            es.shutdown();
 
             if (result == 0) {
                 String commandOutput = out.get();


core/src/main/java/cz/xtf/core/openshift/CLIUtils.java
```diff
@@ -43,7 +44,12 @@ public static String executeCommand(Map<String, String> environmentVariables, St
                 }
             });
 
-            int result = p.waitFor();
+            // Wait for command 10 seconds
+            p.waitFor(10, TimeUnit.SECONDS);
```
simkam, line 48: it could be used for long running actions. I see usages like
```
CLIUtils.executeCommand(podmanCommand(), "build", "-t", IMAGE_NAME, "-f", ...
```
podman build can easily take longer than 10 seconds. There must be some better way to align termination of Process and ExecutorService.

kstekovi: Hi @simkam The `waitFor` is not a root cause. So we can keep it without any timeout. But we have to close threads for Output and Error stream. This will solve my issue.

simkam (approved):

simkam:
thanks @kstekovi!

PR #583, opened 2024-09-18
implement jgit repository resolver

### LINKED ISSUE #574: Dynamically find out git repo of current repo from .git folder
Sounds hacky, but can be useful to avoid unnecesary configuration. We find this can be nice to have in OS TS when pointing helm chart to testsuite repo itself

Also note, similar code is already available in intersmash https://github.com/Intersmash/intersmash/blob/main/core/src/main/java/org/jboss/intersmash/IntersmashDeploymentsGitHelper.java just it needs to be generalized to be used with any repo, not just intersmash.

laDok8:
Introduce `xtf.git.repository.url` and `xtf.git.repository.ref` properties intended for pointing to TS repo itself. 
This should get auto resolved via jgit, but can be overridden by manually setting properties

closes #574

### DIFF
--- a/core/src/main/java/cz/xtf/core/git/JGitResolver.java
+++ b/core/src/main/java/cz/xtf/core/git/JGitResolver.java
@@ -1,0 +1,140 @@
+package cz.xtf.core.git;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.eclipse.jgit.lib.ObjectId;
+import org.eclipse.jgit.lib.Ref;
+import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
+import org.eclipse.jgit.transport.RemoteConfig;
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Try to resolve repository remote URL and branch from .git directory
+ * <p>
+ * This method tries to match HEAD commit to remote references.
+ * If there is a match, the remote URL and branch are set.
+ * For attached HEAD this method could be simplified with jgit methods, but
+ * "universal" approach was chosen, as for example Jenkins git plugin creates a detached state.
+ * In case of multiple matches, upstream or origin (in this order) is preferred.
+ * </p>
+ */
+@Slf4j
+class JGitResolver implements GitResolver {
+    private static final String URL_TEMPLATE = "https://%s/%s/%s";
+    private static String reference;
+    private static String url;
+
+    /**
+     * Try to set repository ref and URL from HEAD commit
+     */
+    public JGitResolver() {
+        try {
+            resolveRepoFromHEAD();
+        } catch (IOException | URISyntaxException e) {
+            log.error("Failed to resolve repository from HEAD", e);
+            throw new RuntimeException("Failed to resolve repository from HEAD with error: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Try to set repository ref and URL from HEAD commit
+     */
+    private static void resolveRepoFromHEAD() throws IOException, URISyntaxException {
+        //look for a git repository recursively till system root folder
+        Repository repository = new FileRepositoryBuilder().findGitDir().build();
+
+        if (repository == null) {
+            log.error("Failed to find a git repository");
+            return;
+        }
+
+        //get current commit hash
+        ObjectId commitId = repository.resolve("HEAD");
+
+        //get all remote references
+        List<Ref> refs = repository.getRefDatabase().getRefs().stream()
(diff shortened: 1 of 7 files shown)


simkam:
is the core really correct location? Wouldn't `test-helpers` module fit better?

laDok8:
I'm not sure but all configurable properties (openshift,helm or XTF itself) seem to be in `cz.xtf.core.config`. Considering current content of `test-helpers` I don't see that many similarities

mnovak1:
Looks like maven dependency `org.eclipse.jgit` is JDK 11 but XTF is based on JDK 8.

mchoma:
Can we have run from OS TS showing this dynamic resolver works as expected.  That mean without those properties in configuration

```
# Repo data
xtf.git.repository.url=<URL>
xtf.git.repository.ref=<BRANCH>
```

core/src/main/java/cz/xtf/core/git/GitRemoteResolver.java
```diff
@@ -0,0 +1,140 @@
...
+        if (pathTokens.length != 2) {
+            log.info("Unexpected path '{}' in URI '{}' of git remote ref '{}'",
+                    remoteConfig.getURIs().get(0).getPath(), remoteConfig.getURIs().get(0), remoteReference);
+            return null;
+        }
+        // the URI must be in HTTPS format
+        return getRepositoryUrl(remoteConfig.getURIs().get(0).getHost(), pathTokens[0], pathTokens[1]);
+    }
+
+    static {
```
mnovak1, line 125: I'm not in favor to have this executed when it's not needed. Can the whole class be written as a singleton. It would be used like: 
``` 
GitRemoteResolver gitRemoteResolver = GitRemoteResolver.getInstance(); // all initialization happens here
gitRemoteResolver.getRepositoryReference();
gitRemoteResolver.getRepositoryUrl();
```

mnovak1 (changes requested):

core/src/main/java/cz/xtf/core/git/GitRemoteResolver.java
laDok8: This code is not always executed, it's executed after loading to memory, which according to my testing happens upon first method call of mentioned class

mnovak1: Ok, still I think that using Singleton would provide cleaner design.

core/src/main/java/cz/xtf/core/git/GitRemoteResolver.java
```diff
@@ -0,0 +1,140 @@
...
+    private static String getRepositoryUrl(String host, String remote, String repository) {
+        return String.format(URL_TEMPLATE, host, remote, repository);
+    }
+
+    /**
+     * Try to set repository ref and URL from HEAD commit
+     */
+    private static void resolveRepoFromHEAD() throws IOException, URISyntaxException {
+        //look for a git repository recursively till system root folder
+        Repository repository = new FileRepositoryBuilder().findGitDir().build();
```
honza-kasik, line 46: Would it make sense to use `findGitDir(File current)` instead and make `current` configurable?

laDok8: IMO not really, at this point user could specify `xtf.git.repository.url` which is also more universal than "hardcoding" local path

core/src/main/java/cz/xtf/core/git/GitRemoteResolver.java
laDok8: now the methods in Resolver are private, this would just move code you mentioned into `GitRemoteConfig` and would result in no public facing API change

mnovak1: There would be no public facing change but internal code could use Singleton, Factory, or Strategy pattern to make it more maintanable. However I'm ok to let it be if we don't anticipate more complex use cases.

core/src/main/java/cz/xtf/core/git/JGitResolver.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Try to set repository ref and URL from HEAD commit
+     */
+    private static void resolveRepoFromHEAD() throws IOException, URISyntaxException {
+        //look for a git repository recursively till system root folder
+        Repository repository = new FileRepositoryBuilder().findGitDir().build();
+
+        if (repository == null) {
+            log.debug("Failed to find a git repository");
```
kstekovi, line 54: can be these log into error instead of debug.. I would like to see a message if something is wrong. now you have to do extra step to enable debug logging to see what is wrong.

mnovak1 (approved):

mnovak1:
After discussion with @laDok8 the functionality is working. If any other changes are required, those will be handled in a follow up PR. 

Merging.

